PR #26, opened 2019-09-26
Generalise basic features necessary for XRD simulations

dnjohnstone:


dnjohnstone:
I think we may as well get this little function in for 0.2.0

pc494:
Closed as stale (and beyond current package scope)

PR #29, opened 2019-10-11
Starting 0.2.0-rc.1

pc494:
---
name: Starting 0.2.0-rc.1
about: This is the first PR on the long road to a January 2020 release, my aim is to lean the repo up, and to make as much as possible stable (and correct)

---

**Release Notes**
> improvement & bugfix & developer change
Summary: TBC

At the moment mainly deletions and refactors.


dnjohnstone:
I'd quite like to keep the Kirkland scattering factors please - they're a community standard and I think the better solution would be to make them an option in simulations rather than remove them...

pc494:
> I'd quite like to keep the Kirkland scattering factors please - they're a community standard and I think the better solution would be to make them an option in simulations rather than remove them...

I'll write the refactor to make them accessible in the code, it looked very much like they were unused at present.

Issue #11, opened 2019-07-16
save/load functions need overhaul


dnjohnstone:
Saving/loading is currently pretty ugly - only defined for library objects and just done by some pretty basic pickle-ing. Now that diffsims is separate from pyxem it's more obvious that this needs to be addressed properly.

pc494:
Unless somebody has very good solution that will be easy to make work I'm going to push this to `0.3.0`

dnjohnstone:
I think the solution here should be to create an hdf5 save/load interface following a pattern similar to https://github.com/pyxem/orix/pull/67

Does that sound ok to others?

hakonanes:
For this I would recommend using functions similar to `hdf5group2dict()` (https://github.com/pyxem/orix/blob/master/orix/io/plugins/orix_hdf5.py#L59) and `dict2hdf5group()` (https://github.com/pyxem/orix/blob/master/orix/io/plugins/orix_hdf5.py#L334). They have been adapted from HyperSpy's HDF5 plugin (hspy: https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/hspy.py#L405, https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/hyperspy/io_plugins/hspy.py#L590). By passing `recursive=True` to `hdf5group2dict()`, one can read datasets within groups within groups... into nested dictionaries. Likewise, nested dictionaries can be written to datasets within groups within groups... within an opened HDF5 group/file. This makes creating HDF5 IO quite OK.

hakonanes:
Is there any interest in this? I haven't seen any. I suggest to close, but re-open whenever the need arises.

pc494:
agreed, closing.

PR #30, opened 2019-10-17
Starting 0.2.0 rc.1 (Supersedes #29)

pc494:
---

name: Starting 0.2.0 rc.1 (Supersedes #29)
about: As previous, with details to follow.

---

**Release Notes**
> major 
> improvement & bugfix & developer change

**Work in progress?**
(more to be added)
- [ ] ~Kirkland Scattering~ (move to an issue, and do in a separate PR)
- [ ] PR template
- [ ] debeye waller docstring is slightly wrong (should be `dict {str:float} or float`)


dnjohnstone:
I think we need to address #11 in this one to sort saving/loading of libraries etc.

Probably using an hdf5 format is the best route - @stefsmeets I know we discussed this in the past, do you have any opinions / suggestions for this?

stefsmeets:
> Probably using an hdf5 format is the best route - @stefsmeets I know we discussed this in the past, do you have any opinions / suggestions for this?

I agree hdf5 would be most flexible. I tried pickling (also using `dill`), but it crashed out on me. But I'm worried that hdf5 will be difficult to maintain, because a lot of custom attributes are being stored on different objects (also on `pyxem`). Hyperspy's `BaseSignal` already has a `.save` method, can we not make use of that somehow?

dnjohnstone:
Yeah, in pyxem we intend to mostly use the hyperspy.save / load method that already exists.

In the case of simulated libraries though, these things are generated by diffsims, which we're trying to keep a lighter dependency list for. i.e. it does not depend on hyperspy and so we're going to have to do something different...

dnjohnstone:
Is this PR dead @pc494 ?

pc494:
Yep, I'll pull out the useful stuff somewhere else.

PR #31, opened 2019-10-18
Minor tweaks

stefsmeets:
Some minor tweaks to the code I made, it fixes a missspelling of 'orthorhombic', and refactors some code for better readability.

### DIFF
--- a/diffsims/generators/library_generator.py
+++ b/diffsims/generators/library_generator.py
@@ -131,6 +131,74 @@ class DiffractionLibraryGenerator:
         return diffraction_library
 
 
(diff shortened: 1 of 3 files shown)


stefsmeets:
My bad, I thought I had fixed this.

dnjohnstone:
Thanks @stefsmeets - we need a test for _generate_lookup_table when unique=False

That seems to be new functionality added here? I like that but just curious as to why you added it?

Happy to merge when that test is done.

stefsmeets:
> I like that but just curious as to why you added it?

I was trying to understand this code so I was testing a few things, and it seemed appropriate to leave it in. :-)

Issue #16, opened 2019-07-16
Add FFT based kinematical simulator


dnjohnstone:
Simulations using an FFT based kinematical simulator would be useful.

PR #33, opened 2019-11-05
FFT based kinematic simulator

robtovey:
I have added a new branch called, for lack of a better name, 'kinematic_simulator_rob' which provides an FFT based simulation of diffraction images which follows an Ewald Sphere model. This hopes to address the enhancement request of #16. Coverage is incomplete but I will continue to add as other issues are dealt with. There is a small hole in the functionality in that I have only implemented a test version of a probe function but I will try to correct that in a minute...

---
Broad structure of the new code:
- generators.diffraction_generator has a new object AtomicDiffractionGenerator which is intended as the front-end interface
- sims.kinematic_simulation provides code which, essentially, converts a discretised volume into a diffraction pattern
- utils.discretise_utils provides code (cpu and cuda) for discretising a volume with atoms inside it
- utils.fourier_transform provides some wrappers of various fft codes, automatically selecting pyfftw if available
---
Some house-keeping questions:
- [x] I have assumed all units are angstroms, do I need to convert at any point?
- [x] Documentation in diffsims.diffraction_generator suggests that energy has unity kV, should this be keV?
- [x] In the code I adapted from I defined my floating precision basically globally but I'm not sure what default to use here. Could it be assumed that everyone uses float32 (or float64)? If not, I think either it should be a user parameter or automatically guessed from one of the inputs. 
- [x] Current convention is a Fourier transform with no factor of pi. Is there a global convention for this or should it be a user preference?
- [x] When discretising an atom there is a cut-off radius which, currently just rounds anything smaller than e^{-20} to 0. Again, should this be a user parameter?
- [x] Is there a current convention for mesh vectors? i.e. if `x, y` are the two coordinate vectors then is `x == x.reshape(-1)` or `x == x.reshape(-1, 1)` or something else?
---
Other things I need to do:
- [x] Implement a generic probe class
- [x] Add appropriate documentation when a probe is an input parameter

p.s. I apologise for my editor's rather religious attitude when it comes to upholding pep8's white-space conventions.


robtovey:
Hopefully this is now in a stable state. I think the only things left are to finish the tests and include a demo notebook. Hopefully I'll get around to these by next week.

dnjohnstone:
@robtovey do you need anything from us to move this forward?

robtovey:
@dnjohnstone No, sorry, I'm just being slow to get around to things on my end.

robtovey:
I think this is basically fully covered now. There are three big holes left
- the GPU code is not tested by travis
- there are several blocks of code designed for anticipating memory errors
- numba code does not register for coverage, this makes sense because it compiles the lines differently

What would you like me to do with the remaining bits? I would suggest excluding them from coverage with a comment to say why in each instance?

Otherwise, two remaining tasks are to make a demo and add the copyright etc.

pc494:
_1) the GPU code is not tested by travis_
That's fine by me.

_2) there are several blocks of code designed for anticipating memory errors_
Could you give a bit more detail on this?

_3) numba code does not register for coverage, this makes sense because it compiles the lines differently_
When you say "does not register" you mean that it claims not to be covered when it is?

_What would you like me to do with the remaining bits? I would suggest excluding them from coverage with a comment to say why in each instance?_
That sounds like a good plan.

_Otherwise, two remaining tasks are to make a demo and add the copyright etc._
Yep, at this point can you just talk about how you expect this code to install/work on machines that both are and aren't equipped with a GPU. It seems that if travis passes you have a fall back for non-GPU work?

robtovey:
> _1) the GPU code is not tested by travis_
> That's fine by me.
> 
> _2) there are several blocks of code designed for anticipating memory errors_
> Could you give a bit more detail on this?

So the main idea is you want to have a high-resolution discretisation but the full arrays can't all be stored on the GPU at once. In which case the arrays are iteratively partitioned until they do fit onto the GPU. There's a little more to it and it will do the same thing even if you're running on the CPU and running out of RAM but this is the main reason and target for it.
> 
> _3) numba code does not register for coverage, this makes sense because it compiles the lines differently_
> When you say "does not register" you mean that it claims not to be covered when it is?

Exactly, I know the functions are called but travis doesn't register any of the lines having been run. This is even true, the actual lines are parsed and compiled by numba but never run by python itself...
> 
> _What would you like me to do with the remaining bits? I would suggest excluding them from coverage with a comment to say why in each instance?_
> That sounds like a good plan.

Great, I'll start filling that in.
> 
> _Otherwise, two remaining tasks are to make a demo and add the copyright etc._
> Yep, at this point can you just talk about how you expect this code to install/work on machines that both are and aren't equipped with a GPU. It seems that if travis passes you have a fall back for non-GPU work?

Sure, functionality is the same between the two. Both are (hopefully) trivial to install on anaconda and a bit more effort on pip but still possible. They should look something like:
- CPU: conda install numba diffsims
- GPU: conda install cudatoolkit numba diffsims
If there is a keyword for using the GPU, on by default, which is silently ignored if cudatoolkit is not installed or if the GPU doesn't activate for some reason.

dnjohnstone:
@robtovey what's left to do on this? It would be good to get merged up before Christmas!

robtovey:
I think we're pretty close, I now have a demo notebook. I'm not sure quite sure where to put it and I might need a bit of advice on the wording but it does the job. The images are not a great resolution but it all runs in 2 minutes on my CPU so hopefully that tradeoff is appreciated.

Something I thought about, do you want code for computing the volume charge density? I've relatively 'hacked' it in as you will see but should I give it a quick polish too?
Also, I have currently only copied out the data for the first 21 elements of the periodic table... How many would you like?!

Other than that, I think the only remaining thing for me is to test the GPU code as well the we can merge.

robtovey:
Hopefully that is finally back to '100%' coverage...
The GPU passed all my tests locally as well so that should be fine.
For convenience, I have put the notebook [here](https://github.com/robtovey/diffsims/blob/kinematic_simulator-rob/Diffsims%20simulate%20atomic%20diffraction%20pattern.ipynb) for now. I should probably try to improve the images before it's merged too.

On another note, not sure if it was deliberate or slipped through the cracks but I think there is only 1 line missing from coverage now in the file: [here](https://coveralls.io/builds/27557274/source?filename=diffsims/libraries/diffraction_library.py#L138)

pc494:
I am looking to tidy this up and get it merged soon(ish). @robtovey - do you think this is "done" as far as you are concerned?

Diffsims simulate atomic diffraction pattern.ipynb
```diff
@@ -0,0 +1,1985 @@
+{
```
dnjohnstone, line 1: Please can we remove this notebook from here and create a pyxem/diffsims-demos repository for tutorials like this.

diffsims/__init__.py
```diff
@@ -22,12 +22,16 @@
 
 import numpy as np
 
-from .generators.diffraction_generator import DiffractionGenerator
+from .generators.diffraction_generator import DiffractionGenerator, AtomicDiffractionGenerator
```
dnjohnstone, line 25: I think we should be able to come up with better names than this - I think the difference is really that we have a structure factor based simulation and an FFT based simulation, we also have (a bit hidden in the code) a real space sum of scattering from many atoms solution. So I think we should either have one DiffractionGenerator with multiple methods or name more based around the simulation approach. I prefer the latter here so maybe we could go with:

SFDiffractionGenerator
FFTDiffractionGenerator
RSDifractionGenerator

Or we could change and have:

StructureFactorSimulator
FFTSimulator
RealSpaceSimulator

What do others think?

diffsims/__init__.py
```diff
@@ -22,12 +22,16 @@
...
 from .generators.library_generator import DiffractionLibraryGenerator
 from .generators.library_generator import VectorLibraryGenerator
 
 from .sims.diffraction_simulation import DiffractionSimulation
 
+from .utils.probe_utils import probeFunction, BesselProbe
+from .utils.fourier_transform import toRecip, fromRecip, getRecipPoints, getDFT
+from .utils.discretise_utils import getDiscretisation
+
```
dnjohnstone, line 34: For all 3 of the above lines - we try to keep to lower case for functions and use only capitals like that in class names. This is just a code style thing, but consistency on that is helpful.

diffsims/generators/diffraction_generator.py
```diff
@@ -201,7 +202,7 @@ def calculate_profile_data(self, structure,
             d_hkl = 1 / g_hkl
 
             # Bragg condition
-            #theta = asin(wavelength * g_hkl / 2)
+            # theta = asin(wavelength * g_hkl / 2)
```
dnjohnstone, line 205: If we're not using this, let's remove it.

diffsims/generators/diffraction_generator.py
```diff
@@ -228,11 +229,11 @@ def calculate_profile_data(self, structure,
             # Intensity for hkl is modulus square of structure factor.
             i_hkl = (f_hkl * f_hkl.conjugate()).real
 
-            #two_theta = degrees(2 * theta)
+            # two_theta = degrees(2 * theta)
```
dnjohnstone, line 232: Likewise, remove if unused.

diffsims/sims/kinematic_simulation.py
```diff
@@ -0,0 +1,187 @@
+"""
+Created on 1 Nov 2019
+
+Back end for computing diffraction patterns with a kinematic model.
+
+@author: Rob Tovey
+"""
```
dnjohnstone, line 25: We need to put the standard copyright notice at the top of all files in this PR.

diffsims/sims/kinematic_simulation.py
```diff
@@ -0,0 +1,187 @@
...
+    pointwise : bool
+        Optional parameter whether atomic intensities are computed point-wise at
+        the centre of a voxel or an integral over the voxel. default=False
+
+    Returns
+    -------
+    DP : ndarray
+        The two-dimensional diffraction pattern evaluated on the reciprocal grid
+        corresponding to the first two vectors of <x>.
+    """
```
dnjohnstone, line 73: I think (may be misremembering) that this doesn't comply with the numpy docstring standards - isn't it after the : it should just be the type and details regarding shape etc should all be in the next line?

dnjohnstone (commented):
I've made some comments mostly just on naming and naming style - otherwise it basically looks good. I will make a diffsims-demos repository now to put the notebook in.

dnjohnstone:
diffsims-demos - now here: https://github.com/pyxem/diffsims-demos/

diffsims/__init__.py
robtovey: I agree with the latter. The main problem I see with trying to merge them into one object is that, as far as I understand it, they just take different inputs (whole crystal vs unit cell) and give different outputs (single diffraction pattern vs complete list of Bragg angles).

I'm not entirely sure what your 'real space' simulator is but I'm not sure about using FFT as part of the naming system. If you add a multi-slice implementation later then I think that will also use FFTs but I guess you'd also want to give it another name? I think these are all fundamentally different models and a choice of backend should be a different parameter/method.

I'm not sure what would be a better name, I guess you will both know much better than me. FirstBorn, Ewald, or something with Kinematic?.. Other than that, I don't have much preference over the general naming format.

diffsims/sims/kinematic_simulation.py
robtovey: Sorry, I wasn't following any formal standards, I'll go back and do that.

diffsims/sims/kinematic_simulation.py
robtovey: I'll do this too

robtovey:
I think this is basically ready to merge now. Two outstanding comments, the naming of diffraction generators and moving the jupyter notebook to its new home. I think the second is easier for either of you to do and once you've decided on a naming system I'll be happy to implement?

robtovey:
Sorry, I'm not sure why appveyor is failing. It's complaining about an import of `matplotlib` which isn't installed but as far as I can tell, this isn't in my code. Have I missed something?

pc494:
The appveyor thing is unrelated to you @robtovey and I'll (attempt to) deal with it, then get this PR sorted and merge for you.

robtovey:
I've just remembered to talk about installation. The minimum are in the files but there are a couple of packages which will accelerate things, in particular using the GPU and fftw backends.
A full installation would include something like:
`conda install cudatoolkit, pyfftw`

PR #41, opened 2020-01-09
Attempting to fix Apveyor

pc494:
---
name: Attempting to fix Apveyor
about: Puts the dependencies (as defined in `setup.py` into `DEPS`)

---

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,7 @@
 environment:
 
   global:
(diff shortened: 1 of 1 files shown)


dnjohnstone:
What was wrong with 0074756 ?

pc494:
> What was wrong with [0074756](https://github.com/pyxem/diffsims/commit/0074756bd4c7b3291c983c0ec8e2e4ffe9a7a87a) ?

I wanted to have the more natural `pip install .` to run without pre-installs, but it isn't essential, will close and reopen a new one with the fix only.

Issue #52, opened 2020-01-22
Improving streographic triangle functionality


pc494:
Our current `streographic_triangle_implementation` was designed to work with an older pyxem architecture. To update it I propose the following.

Create a function `get_beam_direction_streographic(density,crystal_system) ` 
which more or less replicates (with bugfixes) the existing functionality.

Add a second function, `get_rotation_list_streographic(density,beam_resolution,crystal_system)` that behaves in a manner that is consistent with other `rotation_list` generators.

pc494:
#61 closes this.

PR #55, opened 2020-01-23
Rotation lists from streographic gridding

pc494:
---
name: Rotation lists from streographic gridding
about: As is industry standard we will provide an implementation of rotation list generation that use streographic triangles. See #52 

- [ ] ready for review and merge?
---

**Release Notes**
> major
> improvement / bugfix
Summary: Overhaul of streographic functionality

At a user level this provides a `rotation_list` that is based on the streographic list, at an internal level this is done by first getting beam directions, and then applying rotations about them, using the `'szxz'` trick we tried for `get_grid_around_beam_direction`, with adjustment illustrated by #53. Theoretically it abuses spherical polar coordinates, as most reduced regions are defined by lines of "longtitude" and the equator, which puts hard limits on theta and phi. The exception is the cubic case.

**Work in progress?**
- [x] Needs better tests
- [x] I need to decide if equal area is worth doing
- [x] I need to fix the cubic case, which currently samples slightly too wide a region
- [ ] I need to write the theory out somewhere
- [ ] Delete old code that did this
- [ ] Make `StructureLibrary` compliant with all these changes

** Script for checking IPZ figures ** 
Making use of the headline formulae from https://en.wikipedia.org/wiki/Stereographic_projection#Visualization_of_lines_and_planes

```
from diffsims.utils.gridding_utils import get_beam_directions
import numpy as np
from matplotlib import pyplot as plt

xyz = get_beam_directions('cubic',5,equal='area')
x = xyz[:,0]
y = xyz[:,1]
z = xyz[:,2]

xs = np.divide(x,1-z)
ys = np.divide(y,1-z)

plt.scatter(xs,ys)
```

### DIFF
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -22,6 +22,7 @@ Provides users with a range of gridding functions
 
 import numpy as np
 import warnings
+from itertools import product
 
 from diffsims.utils.rotation_conversion_utils import Euler
 from diffsims.utils.fundemental_zone_utils import get_proper_point_group_string, reduce_to_fundemental_zone
@@ -78,7 +79,6 @@ def get_grid_streographic(crystal_system,resolution,equal='angle'):
     -------
     rotation_list : list of tuples
     """
-    from itertools import product
     beam_directions_rzxz = beam_directions_to_euler_angles(get_beam_directions(crystal_system,resolution,equal=equal))
     beam_directions_szxz = beam_directions_rzxz.to_AxAngle().to_Euler(axis_convention='szxz') # convert to high speed convention
 
@@ -149,7 +149,6 @@ def get_grid_around_beam_direction(beam_direction,resolution, angular_range=(0, 
     -------
     rotation_list : list of tuples
     """
-    from itertools import product
 
     if not cubic:
         warnings.warn("This code only works for cubic systems at present")
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -32,7 +32,8 @@ from diffsims.utils.gridding_utils import create_linearly_spaced_array_in_rzxz,r
 
 
 def _returnable_eulers_from_axangle(grid,axis_convention,round_to):
-    """ Converts a grid of orientations in axis-angle space to Euler angles following a user specified convention and rounding."""
+    """ Converts a grid of orientations in axis-angle space to Euler
+    angles following a user specified convention and rounding."""
     eulers = grid.to_Euler(axis_convention=axis_convention)
     rotation_list = eulers.to_rotation_list(round_to=round_to)
     return rotation_list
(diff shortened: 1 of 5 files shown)


pc494:
I'm planning completing the other unticked boxes in a separate PR, which means this is ready for review.

dnjohnstone:
> I'm planning completing the other unticked boxes in a separate PR, which means this is ready for review.

Could you have a look at the broken tests to start, please? Meanwhile I'll have a look at the rest of it

diffsims/generators/rotation_list_generators.py
```diff
@@ -58,6 +58,44 @@ def get_fundemental_zone_grid(space_group_number, resolution):
...
+    equal : 'angle' or 'area'
+        See docstrings for diffsims.utils.gridding_utils.get_beam_directions
+        
+    Returns
+    -------
+    rotation_list : 
+    """
+    from itertools import product
+    beam_directions_rzxz = beam_directions_to_euler_angles(get_beam_directions(crystal_system,resolution,equal=equal))
+    beam_directions_szxz = beam_directions_rzxz.to_AxAngle().to_Euler(axis_convention='szxz') # convert to high speed convention
```
dnjohnstone, line 88: Do we risk any loss of precision with all these conversions? I doubt it, but just thought I'd check.

diffsims/utils/gridding_utils.py
```diff
@@ -169,3 +178,98 @@ def _create_advanced_linearly_spaced_array_in_rzxz(resolution, max_alpha, max_be
...
+    of the streographic triangle are included. Actual 'resolution' will always be equal to or higher than the input resolution. As 
+    an example, if resolution is set to 4 to cover a range [0,90] we can't include both endpoints. The code makes 23 steps
+    of 3.91 degrees instead.
+    
+    For the cubic case: Each edge of the streographic triangle will behave as expected. The region above the (1,0,1), (1,1,1) edge
+    will (for implementation reasons) be slightly more densly packed than the wider region.
+    """
+    theta_max,psi_max,psi_min = crystal_system_dictionary[crystal_system]
+
+    steps_theta = int(np.ceil((theta_max - 0)/resolution)) #see docstrings for np.arange, np.linspace has better endpoint handling
```
dnjohnstone, line 220: is this comment in the right place? linspace used lower down?

dnjohnstone (changes requested):
Thanks @pc494 this looks pretty good - but like you got bored around docstring time.

I think a few more comments in the code might be helpful too, but not essential.

This does need a pepstorm - but because that might affect other parts of the code and make review tricky let's leave that for a separate pepstorm only PR.

diffsims/generators/rotation_list_generators.py
pc494: https://github.com/pyxem/diffsims/blob/746efbc23a9f4210b498cdc2d923b60f26264bcd/diffsims/tests/test_utils/test_rotation_conversion_utils.py#L82

Suggests we can go from euler to axangle and axangle to euler within `1e-8` so I'm not very concerned about this.

diffsims/tests/test_generators/test_rotation_list_generator.py
```diff
@@ -41,6 +41,14 @@ def test_get_grid_around_beam_direction():
 def test_get_fundemental_zone_grid(space_group_number):
     grid = get_fundemental_zone_grid(space_group_number,resolution=3)
 
+def test_get_grid_streographic():
+    grid = get_grid_streographic('tetragonal',1,equal='angle')
+    assert (0,0,0) in grid
+    grid_four_times_as_many = get_grid_streographic('orthorhombic',1,equal='angle')
```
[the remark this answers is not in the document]
pc494: I think the testing of `get_grid_beam_directions` covers everything you could really want here.

diffsims/utils/gridding_utils.py
pc494: I think this is the right place. The point is the "natural" way to do this would be `np.arange(lower_bound,upper_bound,step_size)` but we instead convert our "step_size" to a "num" (on the line with the comment).

diffsims/generators/rotation_list_generators.py
```diff
@@ -58,6 +58,44 @@ def get_fundemental_zone_grid(space_group_number, resolution):
     fz_grid_axangle = reduce_to_fundemental_zone(raw_grid_axangle, zone_string)
     return _returnable_eulers_from_axangle(fz_grid_axangle,'rzxz',round_to=2)
 
+def get_grid_streographic(crystal_system,resolution,equal='angle'):
+    """
+    Creates a rotation list by combining the minimum region of the streogram's beam directions
+    with in plane rotations
```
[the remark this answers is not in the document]
pc494: I think this would be fine if we dropped "and for a specified crystal symmetry" at the end. We don't use the symmetry of the crystal for the inplane rotations.

dnjohnstone (approved):

PR #62, opened 2020-01-30
Feature - ZAP map

### LINKED ISSUE #51: Around "beam direction" functionality
This stems from a discussion on #47 

Two types of functionality that involve user input "beam directions" are desired for diffsims.

1) A way to take a known structure and generate zone axis patterns (ZAPs) #7 
2) The generation of a rotation list about a certain beam direction

In (1) the user will have to provide a diffpy structure, this includes a metric tensor (or at least the parts needed to create one) which in turn allows for calculations of appropriate euler angles

In (2) no such structure needs to be provided (and nor should it), but without that simply saying "I want to be down a [110]" isn't sufficient, the rotation to a cubic [110] is not the same as the one to an orthorhombic [110].

**My proposed solution**

- Make `get_grid_around_beam_direction()` take an euler triplet as it's main argument.
- Write a util that takes a structure and a desired beam direction and returns an euler triplet eg:
`get_rotation_to_beam_direction(structure,beam_direction)`
that takes into account the crystal metric tensor
- Use the above util for a ZAP map function

pc494:
---
name: Feature - ZAP map
about: Addressing #7 and adding functionality that will resolve #51 

---

**Release Notes**
> major 
> new feature
Summary: ZAP map 
get_grid_around_beam_direction 

This is a **WIP** that will eventually provide:
An automated method for inspecting zone axis patterns
In doing so create the tools needed to make `get_grid_around_beam_direction` smooth

Feedback on exactly what should be input/output would be good, currently I'm expecting to return a dictionary that can then be plotted (or inserted into a plotting function if needs be). Is this a good design?

### DIFF
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -24,6 +24,8 @@ import numpy as np
 import warnings
 from itertools import product
 
+from transforms3d.euler import euler2axangle,axangle2euler
+
 from diffsims.utils.rotation_conversion_utils import Euler
 from diffsims.utils.fundemental_zone_utils import get_proper_point_group_string, reduce_to_fundemental_zone
 from diffsims.utils.gridding_utils import create_linearly_spaced_array_in_rzxz,rotate_axangle, \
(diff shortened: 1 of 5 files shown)


dnjohnstone (approved):
Thanks @pc494 - this looks good to me, the only thing I'd question is whether or not this should be a Generator class given it takes a structure and other info. But I'm not so bothered about that and will merge this as is to move forwards.

PR #77, opened 2020-02-21
Function: Get StructureLibrary size (total # of orientations)

norskie:
---
name: Get StructureLibrary size
about: Adds a function to the class StructureLibrary that returns the total number of orientations
---

Adds a function to the class StructureLibrary that returns the total number of orientations and the possibility to print each identifier with its corresponding number of orientations(setting to_print = True). This function may be used to monitor the the library size, when for example doing Template Matching.

Example output when argument to_print is True.
```
Al has 720 number of entries.
Betap has 360 number of entries.
BetaDP has 360 number of entries.
There are in total 1440 number of entries
```
The function returns:
`1440`

### DIFF
--- a/diffsims/libraries/structure_library.py
+++ b/diffsims/libraries/structure_library.py
@@ -56,6 +56,35 @@ class StructureLibrary():
         for ident, struct, ori in zip(identifiers, structures, orientations):
             self.struct_lib[ident] = (struct, ori)
 
+    def get_library_size(cls, to_print = False):
+        """
+        Returns the number of structures in the current StructureLibrary
+        object. Will also print the number of entries for each structure in
+        the library if the parameter to_print is set to True. Works with both
+        cls.orientations on the form [1,2] and [[1,2], [2,3]](rotation lists).
+
+        Parameters
+        ----------
+        to_print : bool
+            Default is 'False'
+            Returns
+            -------
+            size_library : int
+                Total number of entries in the current StructureLibrary object.
+        """
+        size_library = 0
+        for i in range (len(cls.orientations)):
+            if type(cls.orientations[i]) != list:
+                size_library += 1
+            else:
+                size_library += len(cls.orientations[i])
+            if to_print == True and type(cls.orientations[i]) == list:
+                    print(cls.identifiers[i], "has", \
+                    len(cls.orientations[i]), "number of entries.")
+                    if to_print == True:
+                        print("\nIn total:", size_library, "number of entries")
+                    return size_library
+
     @classmethod
     def from_orientation_lists(cls, identifiers, structures, orientations):
         """
--- a/diffsims/libraries/structure_library.py
+++ b/diffsims/libraries/structure_library.py
@@ -105,7 +105,7 @@ class StructureLibrary():
             orientations.append(get_grid_streographic(system, resolution, equal))
         return cls(identifiers, structures, orientations)
     
-     def get_library_size(self, to_print = False):
+    def get_library_size(self, to_print = False):
         """
         Returns the the total number of orientations in the
         current StructureLibrary object. Will also print the number of orientations
--- a/diffsims/libraries/structure_library.py
+++ b/diffsims/libraries/structure_library.py
@@ -56,12 +56,12 @@ class StructureLibrary():
         for ident, struct, ori in zip(identifiers, structures, orientations):
             self.struct_lib[ident] = (struct, ori)
 
-    def get_library_size(cls, to_print = False):
+    def get_library_size(self, to_print = False):
         """
         Returns the the total number of orientations in the
         current StructureLibrary object. Will also print the number of orientations
         for each identifier in the library if the parameter to_print is set to
-        True. Works with both cls.orientations on the form [1,2]
+        True. Works with both self.orientations on the form [1,2]
         and [[1,2], [2,3]](rotation lists).
 
         Parameters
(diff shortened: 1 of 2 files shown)


diffsims/libraries/structure_library.py
```diff
@@ -56,6 +56,36 @@ def __init__(self,
         for ident, struct, ori in zip(identifiers, structures, orientations):
             self.struct_lib[ident] = (struct, ori)
 
+    def get_library_size(cls, to_print = False):
```
pc494: This is a "method" rather than a "class method". It's probably more jargon than you need, but you should have `self `rather than `cls`

diffsims/libraries/structure_library.py
```diff
@@ -56,6 +56,36 @@ def __init__(self,
...
+        to_print : bool
+            Default is 'False'
+            Returns
+            -------
+        size_library : int
+            Total number of entries in the current StructureLibrary object.
+        """
+        size_library = 0
+        for i in range (len(cls.orientations)):
+            if type(cls.orientations[i]) != list:
```
pc494: Can you explain the case you think this catches?

diffsims/tests/test_library/test_structure_library.py
```diff
@@ -21,6 +21,17 @@
 
 from diffsims.libraries.structure_library import StructureLibrary
 
+def test_get_library_size():
+    identifiers = ['a', 'b']
+    structures = [1, 2]
+    rotation_list = [[3,1], [4,5]]
```
pc494: These should be `[[(0,0,0),(0,1,0)],[(0,1,0),(0,2,0)]]` or similar

diffsims/tests/test_library/test_structure_library.py
```diff
@@ -21,6 +21,17 @@
...
+    orientations = [1, 2]
```
pc494: This wouldn't be a legitimate input. Has to be a list of lists of tuples really.

pc494 (changes requested):
Couple of minor changes, let me know if you need further clarifications.

diffsims/tests/test_library/test_structure_library.py
dnjohnstone: @norskie do you understand what the issue is here?

norskie: Yes, I think so. rotation_list should have inputs on the form [(3,1),(4,5)]?

diffsims/libraries/structure_library.py
norskie: Thank you! I was unsure about this, I am used to C++, so some of the Python-class stuff is new to me :)

diffsims/tests/test_library/test_structure_library.py
pc494: See my review comment above. Each phase has an associated list of tuples. So you're looking at something like `[[(0,0,0),(0,1,0)],[(0,1,0),(0,2,0)]]`

diffsims/libraries/structure_library.py
norskie: So, in some cases like in "test_from_orientations_method" we have :
`orientations = [3, 4]`
which gives 
`type(self.orientations[i]) != tuple `(list is wrong, as you pointed out earlier)

in the above case, we know that each identifier has a unique orientation, and size_library can be incremented by a flat 1.

In the case where we use a rotation_list for each identifier we have e.q.
`orientations = [(3,4), (4,4,6,7,9)] `
which gives
`type(self.orientations[i]) == tuple `

In the latter case size_library must be incremented by the number of elements in the tuple corresponding to each identifier.

diffsims/tests/test_library/test_structure_library.py
norskie: Changed rotation_list to list of tuples. When you say "these" do you mean that the identifiers and structures should be on another form as well?

diffsims/tests/test_library/test_structure_library.py
norskie: Update: Changed rotation_list to list of tuples

norskie: I see, thank you.

diffsims/tests/test_library/test_structure_library.py
norskie: I see now that _each phase_ should have a list of tuples, I will fix this now.

diffsims/libraries/structure_library.py
pc494: I think this is wrong, but I also think you've figured that out yourself by the looks of your remarks below?

norskie: Now that I have changed rotation list to the form of [[(0,0,0),(0,1,0)],[(0,1,0),(0,2,0)]], I have altered this line of code to check
```
if len(self.orientations[i]) == 1:
                size_library += 1
else:
                size_library += len(self.orientations[i])
```

norskie:
Thank you all for your feedback! I have now attempted to fix the changes requested, please let me know if there is something I missed or did wrong and I will attempt to resolve the problem.

diffsims/tests/test_library/test_structure_library.py
norskie: orientations/rotation_list is now: [[(0, 0, 0), (0.0, 90.0, -180.0)], [(0, 0, 0)]]

pc494:
Will merge this once the tests pass :)

Issue #75, opened 2020-02-15
Probe functions & aberrations


dnjohnstone:
There are some probe function things here, that we should think about supporting: https://github.com/ePSIC-DLS/epsic_tools/pull/15/files

E.g. aberration function:

def FuncAberrUV(u,v,aberrcoeff):

    # this function needs comments

    # input:
    # u: Kx
    # v: Ky

    u2 = u*u
    u3 = u2*u
    u4 = u3*u

    v2 = v*v
    v3 = v2*v
    v4 = v3*v

    # aberr are in unit of meter.
    C1   = aberrcoeff[0] # defocus
    C12a = aberrcoeff[1] # 2 stig
    C12b = aberrcoeff[2] # 2 stig
    C23a = aberrcoeff[3] # 3 stig
    C23b = aberrcoeff[4] # 3 stig
    C21a = aberrcoeff[5] # coma 
    C21b = aberrcoeff[6] # coma
    C3   = aberrcoeff[7] # Spherical abb
    C34a = aberrcoeff[8] # 4 stig
    C34b = aberrcoeff[9] # 4 stig
    C32a = aberrcoeff[10] # star
    C32b = aberrcoeff[11] # star

    # output:  chi function. in unit of meter*radian.  multiply by 2pi/lambda to get dimensionless
    func_aberr =  1/2*C1*(u2+v2)\
            + 1/2*(C12a*(u2-v2) + 2*C12b*u*v)\
            + 1/3*(C23a*(u3-3*u*v2) + C23b*(3*u2*v - v3))\
            + 1/3*(C21a*(u3+u*v2) + C21b*(v3+u2*v))\
            + 1/4* C3*(u4+v4+2*u2*v2)\
            + 1/4* C34a*(u4-6*u2*v2+v4)\
            + 1/4* C34b*(4*u3*v-4*u*v3)\
            + 1/4* C32a*(u4-v4)\
            + 1/4* C32b*(2*u3*v + 2*u*v3)\

    return func_aberr

PR #78, opened 2020-02-21
Probe function simulations

dnjohnstone:
Scoping out structure for adding more complete probe function simulations.

Following discussion with @M0hsend and @chrisallen80 - scoping out the most appropriate place to put probe functions.

This should extend and/or be consistent with the basic probe functions implemented by @robtovey for his FFT based simulations.

Since this bit of the code defines new classes rather than only util functions I think it is more appropriate to put in a new diffsims.probe module as I have done here.

Addresses #75


dnjohnstone:
@chrisallen80 I've basically tried to copy over parts of your code that I can see that seem to relate to simulating a probe function. But it seems like I'm either not finding some parts or they're missing from what I can see E.g. func_ObjApt?

As you can see we already have a generic ProbeFunction class and a specific BesselProbeFunction class - so generalizing this to include lens aberrations and whatever else you had in mind is certainly something we're interested in.

Doing that would be a reasonably possible "small unit of work" - it would be helpful if you could describe here what you were trying to achieve with a probe function to help get everything in the right place.

pc494:
Closed as stale.

PR #82, opened 2020-02-24
[fix] Gimbal locks in _create_rzxz grid

pc494:
---
name: [fix] Gimbal locks in `_create_rzxz grid`
about: Drops gimbal lock rotations from grid creation

---

**Release Notes**
>bugfix 
Summary: Drops gimbal lock rotations from grid creation

Previously our deepest back-end code would generate our array of eulers using product, with 0 included in the beta slot this meant you would get:
```
1 0 1
2 0 0 
etc
```

Which are all the same rotation, this PR fixes this. The product stage doesn't include beta=0 but those rotation are added in a subsequent step as `(0,0,gammas)`


dnjohnstone (commented):
LGTM

PR #85, opened 2020-03-06
Transpose to make consistent with main pyxem repo

pc494:
---
name: Transpose to make consistent with main pyxem repo
about: Currently `.get_diffraction_patterns` produces patterns transposed relative to how they would appear in pyxem. This fixes that problem (will reproduce with example before merging). Code now correctly deals with spots that lie off the detector.

---

**Release Notes**
> improvement
Summary: (Breaking Change)


dnjohnstone:
@pc494 is this still all good? Could we get this one merged up soonish please?

pc494:
yep, had only left it out of `main` so that we could make patch releases without a two branch system

dnjohnstone:
ok - if we're going to dive straight in with a 0.3.0 now, is there any reason not to merge this?

pc494:
I'll fix the conflict tomorrow and merge, that's fine - was leaving it until the build was fixed (thanks, seems that #95 did that), alongside (also now closed #92)

PR #98, opened 2020-07-20
Credits for Rob Tovey

### LINKED ISSUE #97: Credits for Rob Tovey
A good portion of our simulation code comes from @robtovey, but isn't ascribed to him by github yet. It would be nice to fix this.

robtovey:
Pull request to address issue #97.

Original commits in #33 did not log correct details for contribution.
Reverted to original branch then moved forward to final merging commit to link with my account.

Hopefully now it will be easier for people to complain when they find all my bugs ;)


pc494:
Great, thanks for this, I'll get to work on the merge conflicts :)

EDIT: turns out this is slightly more involved than originally anticipated.

dnjohnstone:
I know this is non-ideal - but I do think that we should just accept that a mistake was made and move on, it seems like a lot of effort to resolve for not a lot of benefit?

robtovey:
I had another think about this last night and I don't actually think it will work anyway.

Any changes which I originally made and are still in the master wont conflict so they wont change so they wont gain my reference.
When you go through and solve the conflicts with the master you'll just keep everything the same so similarly, none of the attributions will change on the master.

Overall, this is probably a lot of work for absolutely no change at all...

robtovey:
Sorry, do you think we can confirm which way this is heading?
I don't think this merge will actually do what we were aiming for and I can't think of anything that will.

I'm sorry I made a bit of a mess there but maybe the best thing to do is just let this slide into history...?

pc494:
I think I could make this work, but I've yet to get around to it, if the thing I was planning doesn't get it there I'll give up and, like you say, let it slide into history (if you could leave this PR open that would help though! :) )

robtovey:
That's fine with me. I just hope you don't waste too much time on it.
Feel free to close the PR and issue whenever you want without checking with me first but also ask if there's anything I can help with on my end.

PR #99, opened 2020-07-21
Rob Tovey merge conflicts sorted (carrys #98)

pc494:
This will need a bit of a tidy up afterwards but I'll merge it as is to keep the diffs smaller


pc494:
Nope this doesn't do what we wanted.

PR #104, opened 2020-07-31
Remove gridding

### LINKED ISSUE #100: Migration of grid functionality to orix
Diffsims will now depends on orix.

pc494:
---
name: Remove gridding, closes #100
about: As discussed elsewhere, we are moving the gridding functionality to `orix`, this PR implements that in `diffsims`

---

**In Scope**
- Wrap existing syntax around `orix` ports ~and raise a `DeprecationWarning`~ (FZ gridding and local gridding)
- Delete various utils that supported the above
- Migrate all suitable code into: `rotation_list_generator.py`, for example beam_direction grids (currently in `gridding_utils.py`)
- Removes stereographic gridding (use fundamental zone gridding instead) - maybe with a warning for `v0.3` ?
- ~Any associate tidy ups, including those that relate to #103, (in beam direction form)~
- ~Change how `get_beam_direction` copes with the cubic case to improve brevity~
- Moving (and deleting) the various rotation conversion utils (see https://github.com/pyxem/orix/issues/79)


pc494:
These are going to involve renaming some functions, which we should keep track of. I think the **diffsims** versions will be called `get_xyz_grid_list` and the **orix** version `get_xyz_grid`

dnjohnstone (approved):
Thanks @pc494 always nice to see the maintenance burden drop for a change :)

PR #102, opened 2020-07-29
bugfix for calculate_profile_data

isabelwood100:
Changed the key of the dictionary to unique tuple(hkl) such that it no longer overwrites old entries. This will resolve #32 

Once the new plot is obtained a new bug is observed in get_plot of ProfileSimulation such that it plots all the Miller indices labels for all planes in that family on top of themselves at each Bragg peak rather than plotting the family so i'll look at fixing that.

PR #110, opened 2020-08-21
scope out of scaling function

pc494:
FAO: @JoonatanL 

This is what the "custom scaling function" would look like, the user feeds in a function of the form

```python
def clever_shape_function(excitation_error,maximum_excitation_error,**kwargs):
```
In a (perfect world) we check that the function obeys the obvious physical requirement (that it is 0 when `excitation_error` is bigger than `maximum_excitation_error`). The *kwargs get passed down, so you can tinker as one sees fit.

However, all the code around this bit of code is outdated and in a pretty bad way, so before I move on here (probably in a new PR) we need to work through #102, which is a good start (thanks to @isabelwood100).


pc494:
Superceded by #114, closing.

PR #119, opened 2020-09-07
Sweeping for release: ProfileSimulation and updated dependancies

### LINKED ISSUE #116: g_max argument not used in the get_plot method of DiffractionSimulation
g_max argument not used in the get_plot method of DiffractionSimulation - it should be used or removed.

pc494:
---
name: Sweeping for release, ProfileSimulation, updated dependancies, 
about: Closes #116 by removing the `g_max` argument (if somebody needs it it's an easy fix, and you should aim not to simulate more peaks than you want to plot) and removes the "marked failure" for rotation_lists that relied on the local gridding function. This was fixed in https://github.com/pyxem/orix/pull/118

---

**Release Notes**
>  improvement 
Summary: g_max argument removed from plotting ProfileSimulations


pc494:
**NOTE** ~ This will fail the test until we release orix 0.5.0

hakonanes:
By the way, I'm being told by PyCharm that this line https://github.com/pyxem/diffsims/blob/master/diffsims/generators/diffraction_generator.py#L277 is invalid since `y` is a list which does not define `__truediv__`. Should perhaps fix this in this PR, since it is somewhat related?

pc494:
Yep, will fix that here as it's clearly got a code smell to it.

pc494:
To do:

- [x]  `ProfileSimulation` gmax argument clean up
- [x]  `get_library_entry` clean up 
- [x]   `Fix magnitude_tolerance` in `ProfileSimulation.calculate_ed_data()`
- [x] automate release workflow
- [x] bump version to 0.3.0

PR #122, opened 2020-09-17
Adds a changelog

pc494:
---
name: Adds a changelog
about: Following discussion in orix, we now have a changelog for this package.

---


hakonanes (approved):
Looking good!

Issue #129, opened 2020-09-28
Creating spherical meshes for get_grid_beam_directions


pc494:
**Is your feature request related to a problem? Please describe.**
`get_grid_beam_directions` currently picks a mesh in a rather naive way. As described in [1] several "better" options exist. A fairly extensive discussion can be found in https://github.com/pyxem/diffsims/pull/128. Thanks are due to @din14970 for bringing this to my attention.

**Describe the solution you'd like**
This remains slightly up for discussion. From a programming point of view, I think it's fairly obvious that we want 
`
get_grid_beam_directions(crystal_system,resolution,equal="area")`
to become`
get_grid_beam_directions(crystal_system,resolution,mesh="mesh_type")`

The questions that are then open are 
(a) how many mesh types should we provide?
(b) how should be describe "resolution" for these mesh types

My inclination on (a) is that options 1-3 from the linked article would be enough, and adding an extra package for what at first glance seems a fairly marginal benefit feels wrong to me.

On (b) I think the resolution should probably the worst nearest neighbour distance on the grid, although this is a conservative number. Regardless, we should agree what it is, write it in the docstrings, and use it for all gridding types.

Also, a more formal citation to replace [1] would be good, but that's a bonus.

=====

@din14970 are you happy to be assigned this?

[1] https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4

din14970:
@pc494  yes I can try to handle it. Might need some input on proper unit testing along the way though.

In regards to the remaining questions:

(a) agree 1-3 are a must, 1 is already there as `equal=angle` and 2-3 are just my methods that should be slightly modified. An icosahedral mesh would be nice but since it is generated iteratively it doesn't play nice with the `resolution` argument. One might not need the entire meshzoo package, just copy the relevant code bits for meshing a sphere. They are licensed under GPL-3, this should mean the code can be copied because diffsims is also GPL-3, right? I will experiment with it and see whether it's worth it, then we can discuss in the PR.

(b) yes, I would also opt for largest nearest neighbor distance in degrees. In the UV sphere this is on the equator, in the cube based meshes this is the angle from [001] to its nearest neighbor in the direction of the [111] corner.

Will try to find a proper reference for sphere meshing

pc494:
Sounds good to me :+1:

pc494:
[...]

PR #130, opened 2020-10-05
Improving and enhancing the creation of beam grid

din14970:
---
name: Enhancing beam grid
about: An answer to #129 
---

**Release Notes**
> new features, improvement and bug-fix
Summary:
* changed the way in which `get_beam_directions_grid` works
    * Now accepts the three arguments: `crystal_system`, `resolution`, and `mesh`
    * Mesh accepts 5 options related to how the sphere is meshed: `uv_sphere`, `normalized_cube`, `spherified_cube_corner` (default), `spherified_cube_edge`, and `icosahedral`. The generation of the full sphere mesh is broken up into 3 functions: `get_uv_sphere_mesh_vertices`, `get_cube_mesh_vertices`, `get_icosahedral_mesh_vertices`. The kwargs in `get_cube_mesh_vertices` allows one to create a normalized grid, or a spherified grid in two ways (detailed explanation in docstrings). The icosahedral mesh was copied and adapted from meshzoo. The full mesh is "cropped" as it was before.
   * It is ensured that for all meshes, `resolution` means the maximum angular distance to the first nearest neighbor from any grid point. 
* tests were updated to reflect the change in `get_beam_directions_grid`. I also test that the maximum angle between nearest neighbor points is smaller than resolution over the entire grid.
* minor bug fix in UV sphere mesh (original meshing function) to ensure the maximum angular deviation to the first nearest neighbor is smaller or equal than resolution.
* minor bug fix in `get_beam_directions_grid`: in the "cropping" part I round the dot product to 7 decimals to ensure that numerical errors don't exclude points where the dot product is exactly 0.
* added a few utility functions which are used in the meshing functions:
   * `normalize_vectors` takes a (N,D) array representing N D-dimensional vectors and returns a (N, D) array of those vectors normalized to unit length.
   * `_compose_from_faces` utility function adapted from meshzoo to construct the icosahedral grid
   * `_get_first_nearest_neighbors` from a (N,D) array representing a D-dimensional point cloud of N points, return an (N, D) array representing the first nearest neighbor point of each point
   * `_get_angles_between_nn_gridpoints` if the 3D point cloud can be represented as unit vectors, they are points on a unit sphere and we can calculate the angles between these vectors. The first nearest neighbor of each gridpoint is the vector with the smallest relative angle. This function returns the angles between each grid point and its nearest neighbor. This function could be useful for getting an angular spread distribution for different meshes.
   * `_get_max_grid_angle` get the maximum angle between any point and its nearest neighbor in a sphere grid
   * `_get_min_grid_angle` get the minimum angle between any point and its nearest neighbor in a sphere grid
   * `_vector_grid_to_euler` convert a (N, 3) array representing N 3D vectors to orientations with two euler angles in the bunge convention: (0, Phi, phi2). The `orientation` in this sense means the rotation to turn [001] to be parallel to the vector, with the first euler angle constrained (or the other way around, I always get confused). I use this function to verify my output with MTEX.

Previously, the user had limited control in creating a grid of directions on the fundamental triangle. Only a UV-sphere type grid was possible, which significantly oversamples near the (100) poles in the azimuthal direction. The two options (equal angle versus equal area) were non-sensical. A perfect grid on the sphere is not possible, but more fair options exist, each with their own drawbacks. To give the user maximum control, I added 4 new methods of gridding the sphere (described above). In each case, I ensure that the `resolution` argument is the maximum angle from any point on the grid to its nearest neighbor.
See more detailed descriptions in #129 and #128, as well as https://github.com/pyxem/orix/issues/125

All alternatives I could think of I implemented. One could still go for more exotic sphere meshes but as the resolution is refined, the returns will probably be minimal.

I tried to iron out the issues I could find. One could still verify the corners of the fundamental zone as there were some discussions about this in #128 

**Work in progress?**
- [x] I now implemented everything in the file `rotation_list_generator` for simplicity. However, I can imagine that some helper functions do not belong there and may better fit elsewhere. This I leave up to you. 
- [x] Optional: Testing could always be improved, utility functions are tested implicitly. One should test for the "correctness" of each grid. Now I only test for maximum angle.
- [x] Optional: The icosahedral grid is slow because it is generated iteratively, i.e. a number of starting subdivisions is chosen, the maximum angle on the grid is calculated, if it is larger than resolution, the number of subdivisions is increased. If a maximum angle could be known a priory, the generation of the grid would be much faster.

### DIFF
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -505,14 +505,6 @@ def _get_max_grid_angle(vertices, leaf_size=50):
     return np.max(_get_angles_between_nn_gridpoints(vertices, leaf_size))
 
 
-def _get_min_grid_angle(vertices, leaf_size=50):
-    """
-    Helper function to get the minimum angle between nearest neighbor grid
-    points on a grid.
-    """
-    return np.min(_get_angles_between_nn_gridpoints(vertices, leaf_size))
-
-
 def get_icosahedral_mesh_vertices(resolution):
     """
     Return the (x, y, z) coordinates of the vertices of an icosahedral
@@ -662,7 +654,7 @@ def get_beam_directions_grid(crystal_system, resolution,
     """
     if mesh == "uv_sphere":
         points_in_cartesians = get_uv_sphere_mesh_vertices(resolution)
-    elif mesh == "normalized_cube" or "spherified_cube_edge":
+    elif mesh == "normalized_cube" or mesh == "spherified_cube_edge":
         # special case: hexagon is a very small slice and 001 point can
         # be isolated. Hence we increase resolution to ensure minimum angle.
         if crystal_system == "hexagonal":
@@ -694,7 +686,7 @@ def get_beam_directions_grid(crystal_system, resolution,
         a, b, c = uvtw_to_uvw(a), uvtw_to_uvw(b), uvtw_to_uvw(c)
 
     # eliminates those points that lie outside of the stereographic triangle
-    epsilon = -1e13
+    epsilon = -1e-13
     points_in_cartesians = points_in_cartesians[
         np.dot(np.cross(a, b), c) *
         np.dot(np.cross(a, b), points_in_cartesians.T) >= epsilon
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -196,8 +196,8 @@ def get_uv_sphere_mesh_vertices(resolution):
     ----------
     https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4
     """
-    steps_theta = int(np.ceil(180 / resolution))  # elevation
-    steps_psi = int(np.ceil(360 / resolution))  # azimuthal
+    steps_theta = int(np.ceil(180 / resolution))+1  # elevation
+    steps_psi = int(np.ceil(360 / resolution)) # azimuthal
 
     psi = np.linspace(0, 2 * np.pi, num=steps_psi, endpoint=False)
     theta = np.linspace(0, np.pi, num=steps_theta, endpoint=True)
@@ -612,10 +612,8 @@ def get_beam_directions_grid(crystal_system, resolution,
         'orthorhombic','monoclinic','triclinic'
 
     resolution : float
-        An angle in degrees characterizing the finesse of the grid. Depending
-        on the meshing method there can be a slight variation in the
-        interpretation, but it is intended to mean the worst-case angular
-        distance to a nearest neighbor grid point.
+        An angle in degrees representing the worst-case angular
+        distance to a first nearest neighbor grid point.
 
     mesh : str
         Type of meshing of the sphere that defines how the grid is created.
@@ -630,9 +628,20 @@ def get_beam_directions_grid(crystal_system, resolution,
     if mesh == "uv_sphere":
         points_in_cartesians = get_uv_sphere_mesh_vertices(resolution)
     elif mesh == "normalized_cube":
+        # special case: hexagon is a very small slice and 001 point can
+        # be isolated. Hence we increase resolution to ensure minimum angle.
+        if crystal_system == "hexagonal":
+            resolution = np.rad2deg(
+                    np.arctan(np.tan(np.deg2rad(resolution))/np.sqrt(2)))
         points_in_cartesians = get_cube_mesh_vertices(resolution,
                                                       grid_type="normalized")
+
     elif mesh == "spherified_cube_edge":
+        # special case: hexagon is a very small slice and 001 point can
+        # be isolated.  Hence we increase resolution to ensure minimum angle.
+        if crystal_system == "hexagonal":
+            resolution = np.rad2deg(
+                    np.arctan(np.tan(np.deg2rad(resolution))/np.sqrt(2)))
         points_in_cartesians = get_cube_mesh_vertices(resolution,
                                                       grid_type="spherified_edge")
     elif mesh == "spherified_cube_corner":
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -21,6 +21,8 @@ Provides users with a range of gridding functions
 """
 
 import numpy as np
+from scipy.spatial import cKDTree
+import math
 from itertools import product
 
 from orix.sampling.sample_generators import get_sample_fundamental, get_sample_local
@@ -159,46 +161,50 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0, 
     return rotation_list
 
 
-def get_beam_directions_grid(crystal_system, resolution, equal="angle"):
+def normalize_vectors(vectors):
     """
-    Produces an array of beam directions, evenly (see equal argument) spaced that lie within the streographic
-    triangle of the relevant crystal system.
+    Helper function which returns a list of vectors normalized to length 1 from
+    a 2D array representing a list of 3D vectors
+    """
+    return (vectors.T/np.linalg.norm(vectors, axis=1)).T
+
+
+def get_uv_sphere_mesh_vertices(resolution):
+    """
+    Return the vertices of a UV (polar) mesh on a unit sphere.
+    The mesh vertices are defined by the parametrization:
+
+    x = sin(u)cos(v)
+    y = sin(u)sin(v)
+    z = cos(u)
 
     Parameters
     ----------
-    crystal_system : str
-        Allowed are: 'cubic','hexagonal','trigonal','tetragonal','orthorhombic','monoclinic','triclinic'
-
     resolution : float
-        An angle in degrees. If the 'equal' option is set to 'angle' this is the misorientation between a
-        beam direction and its nearest neighbour(s). For 'equal'=='area' the density of points is as in
-        the equal angle case but each point covers an equal area
-
-    equal : str
-        'angle' (default) or 'area'
+        An angle in degrees. The maximum angle between nearest neighbor
+        grid points. In this mesh this occurs on the equator of the sphere.
+        All elevation grid lines are separated by at most resolution. The
+        step size of u and v are rounded up to get an integer number of
+        elevation and azimuthal grid lines with equal spacing.
 
     Returns
     -------
-    points_in_cartesians : np.array (N,3)
-        Rows are x,y,z where z is the 001 pole direction.
+    points_in_cartesian : np.array (N,3)
+        Rows are x, y, z where z is the 001 pole direction
+
+    References
+    ----------
+    https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4
     """
     steps_theta = int(np.ceil(180 / resolution))  # elevation
     steps_psi = int(np.ceil(360 / resolution))  # azimuthal
 
     psi = np.linspace(0, 2 * np.pi, num=steps_psi, endpoint=False)
-
-    if equal == "angle":
-        theta = np.linspace(0, np.pi, num=steps_theta, endpoint=True)
-
-    elif equal == "area":
-        # http://mathworld.wolfram.com/SpherePointPicking.html
-        v_array = np.linspace(0, 1, num=steps_psi)
-        theta = np.arccos(2 * v_array - 1)  # in radians
+    theta = np.linspace(0, np.pi, num=steps_theta, endpoint=True)
 
     psi_theta = np.asarray(list(product(psi, theta)))
     r = np.ones((psi_theta.shape[0], 1))
     points_in_spherical_polars = np.hstack((r, psi_theta))
-
     # keep only one theta ==0 point, specifically the psi ==0 one
     points_in_spherical_polars = points_in_spherical_polars[
         np.logical_or(
@@ -206,7 +212,6 @@ def get_beam_directions_grid(crystal_system, resolution, equal="angle"):
             points_in_spherical_polars[:, 1] == 0,
         )
     ]
-
     # keep only one theta ==180 point, specifically the psi ==0 one
     points_in_spherical_polars = points_in_spherical_polars[
         np.logical_or(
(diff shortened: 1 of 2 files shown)


pc494:
This looks pretty good to me at a first glance; it'll probably take two go rounds with the review but I'll try and get the first one in at some point this week.

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+        chosen so that the angles between the grid points on the line 
+        connecting the face centers (<001>) to the edges (<011>) are equal.
+        The maximum angle is also between the <001> directions and the first
+        grid point towards the <011> edges.
+
+        The spherified_corner case is similar to the spherified_edge case, but
+        the spacing of the grid lines is chosen so that the angles between
+        the grid points on the line connecting the face centers to the cube
+        corners (<111>) is equal. The maximum angle in this grid is from the
+        corners to the first grid point towards the cube face centers.
```
pc494: can we juggle this a little? keep the parameter descriptions to the one liners and move everything else into a `Notes`. It's a bit dense at the moment

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+    ----------
+    [1] https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4
+    """
+    # the angle between 001 and 011
+    max_angle = np.arccos(1/np.sqrt(2))
+    # the distance on the cube face from 001 to 011 grid point
+    max_dist = 1
+    if grid_type == "normalized":
+        grid_len = np.tan(np.deg2rad(resolution))
+        steps = math.ceil(max_dist/grid_len)
```
pc494: so things like this can go to `np.ceil` (see comment by the import)

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+    -------
+    points_in_cartesian : np.array (N,3)
+        Rows are x, y, z where z is the 001 pole direction
+
+    References
+    ----------
+    [1] https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4
+    """
+    # the angle between 001 and 011
+    max_angle = np.arccos(1/np.sqrt(2))
```
pc494: this could be written explicitly as np.pi/4 or np.deg2rad(45)?

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+                    [1, -1, -1]])
+    # combine
+    all_vecs = np.vstack([bottom, top, east, west, south, north, m_c])
+    return normalize_vectors(all_vecs)
+
+
+def _compose_from_faces(corners, faces, n):
+    """
+    Helper function to refine a grid starting from a platonic solid,
+    adapted from meshzoo [1]
```
pc494: this reference is absent, I suggest removing it here, adding a See Also to the bottom with the main function (which also has the reference)

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+        vertices = _compose_from_faces(corners, faces, n)
+        angle = _get_max_grid_angle(vertices)
+        n = n+1
+    # push all nodes to the sphere
+    norms = np.sqrt(np.einsum("ij,ij->i", vertices, vertices))
+    vertices = (vertices.T / norms.T).T
+    return vertices
+
+
+def _vector_grid_to_euler(vectors):
```
pc494: I think this is unused?

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+        if crystal_system == "hexagonal":
+            resolution = np.rad2deg(
+                    np.arctan(np.tan(np.deg2rad(resolution))/np.sqrt(2)))
+        points_in_cartesians = get_cube_mesh_vertices(resolution,
+                                                      grid_type="spherified_edge")
+    elif mesh == "spherified_cube_corner":
+        points_in_cartesians = get_cube_mesh_vertices(resolution,
+                                                      grid_type="spherified_corner")
+    elif mesh == "icosahedral":
+        points_in_cartesians = get_icosahedral_mesh_vertices(resolution)
```
pc494: can you reorder the logic here to go `if mesh` for the three that are one liners and then:

```
elif mesh == "normalized_cube" or "spherified_cube_edge":
      if crystal_system="hexagonal":
                     # stuff
      if mesh == "normalized_cube":
              points_in_cartesians = ...
     else:
             points_in_cartesians = ...
```

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
     points_in_cartesians = points_in_cartesians[
-        np.dot(np.cross(b, c), a) * np.dot(np.cross(b, c), points_in_cartesians.T) >= 0
+        np.round(np.dot(np.cross(b, c), a) * np.dot(np.cross(b, c),
+                 points_in_cartesians.T), sigfig) >= 0
     ]
     points_in_cartesians = points_in_cartesians[
-        np.dot(np.cross(c, a), b) * np.dot(np.cross(c, a), points_in_cartesians.T) >= 0
+        np.round(np.dot(np.cross(c, a), b) * np.dot(np.cross(c, a),
+                 points_in_cartesians.T), sigfig) >= 0
     ]
```
pc494: Would prefer using `EPSILON = -1e-14` and compare to that, see https://github.com/pyxem/orix/pull/60/files for example

pc494 (changes requested):
This is a good PR, and most of these requests are for/about readability or clarity. I've found in the long term such changes pay off.

diffsims/generators/rotation_list_generators.py
```diff
@@ -21,6 +21,8 @@
 """
 
 import numpy as np
+from scipy.spatial import cKDTree
+import math
```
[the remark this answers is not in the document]
din14970: probably yes, I think i was looking for a `ceil` function and first stumbled on `math`. Will change it. In my opinion, ceil and floor should just be standard python functions.

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
...
+    points_in_cartesian : np.array (N,3)
+        Rows are x, y, z where z is the 001 pole direction
+
+    References
+    ----------
+    https://medium.com/game-dev-daily/four-ways-to-create-a-mesh-for-a-sphere-d7956b825db4
     """
-    steps_theta = int(np.ceil(180 / resolution))  # elevation
-    steps_psi = int(np.ceil(360 / resolution))  # azimuthal
+    steps_theta = int(np.ceil(180 / resolution))+1  # elevation
```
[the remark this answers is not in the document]
din14970: When I was testing for maximum nearest neighbor angle, I found that the elevation gridlines are actually slightly more widely spaced than `resolution`. I think it comes from the lines below these lines where the linspace is generated using or not using the end point. Adding just 1 extra gridline resolved the issue with the tests, but I didn't bother looking more closely into it.

diffsims/generators/rotation_list_generators.py
din14970: Yep completely agree, just wanted to make sure I wrote everything out

diffsims/generators/rotation_list_generators.py
din14970: Not in the other functions, but I used it for converting my grid to euler angles which I export so I could import it to MTEX. I couldn't find a suitable alternative elsewhere, maybe there is already a function doing this? If not, I'm not sure where it fits best, maybe not in this file. If you think it should stay, I would say it's more of a public function.

diffsims/generators/rotation_list_generators.py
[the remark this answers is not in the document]
din14970: I went back just to check and indeed it's because on theta you state `endpoint=True`. So say you want a resolution of 10 degrees, you should have `steps_theta=19` not 18, because you include both endpoints: 0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180 = 19 values. So I'm leaving it with the +1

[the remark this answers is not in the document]
pc494: sounds good to me, glad to have that cleared up.

diffsims/generators/rotation_list_generators.py
```diff
@@ -159,83 +161,521 @@ def get_grid_around_beam_direction(beam_rotation, resolution, angular_range=(0,
     return rotation_list
 
 
-def get_beam_directions_grid(crystal_system, resolution, equal="angle"):
+def normalize_vectors(vectors):
```
hakonanes: This all looks very nice.

Just a quick question: have you considered using `orix.vector.Vector3d` to store vectors (https://orix.readthedocs.io/en/stable/reference.html#module-orix.vector.vector3d)? If `v` is a vector, you can use `v.unit` to get the normalized vectors.

diffsims/generators/rotation_list_generators.py
din14970: For now I have left it in here renamed, made it a public function, added a better docstring, and added a unit test for it. If at some point things get reshuffled and there are better places to put this in feel free to move/delete it.

din14970:
still found some bugs, will fix now

diffsims/generators/rotation_list_generators.py
din14970: Didn't consider it mainly because I wasn't aware of this documentation and because I didn't want to delve too deeply into what is already available in orix. Also, since orix is still changing a lot, it seems safer to implement basic functionality like these meshing functions with lower level things like numpy arrays. I'm not too keen now to look for all the places where I could have used orix, but if you feel like it will improve the maintainability of the code in the long run feel free to change it.

hakonanes: Sounds reasonable, so I wouldn't consider it for this PR.

pc494 (approved):
This looks good to me, I will make a couple of packaging additions and then merge.

pc494:
Sorry @din14970, I have another question. I was looking at `beam_directions_grid_to_euler` and hoping to merge it into `get_rotation_from_z_to_direction` (from `zap_map_generator.py`) but I think there might be a disagreement about euler angle conventions.

diffsims would treat bunge euler angles from left to right and so I would expect `phi2` (the final angle) to be the zero and not `phi1`?

din14970:
@pc494 This also always messes me up and has to do with active vs. passive rotations and also how the euler angles are defined. Actually thanks for making me have a second look at this because it appears what I have is still not quite right.

Just to clarify my solution: 

what I looked at was what I got out of the MTEX plots and what information I can derive from the ASTAR software. So I verified for the cubic case that the rotations I export with my functions do give me back the right IPF plots in MTEX. In the cubic case we want to populate the triangle [0, 0, 1] - [1, 0, 1] - [1, 1, 1]. I verified that this works properly when generating the grid. I then convert to Euler angles with `beam_directions_grid_to_euler`, export, and import to MTEX. I would then expect the full IPF to have this triangle populated with points, because the IPF-Z should tell me which crystal vectors to expect parallel to the world Z-axis in my collection of orientations. The IPF images look like:
image
image
image
image

So while 001 and 111 are ok, I have the wrong 011, which I obviously overlooked because the shape looked ok. Changing `phi2` to `np.pi/2 - phi2` produced the expected triangle:
image

Maybe the right way to do it requires a bit more thought and proper testing.

Regardless, flipping phi1 and phi2 results in the following IPF, which is clearly incorrect:
image

A final bit of evidence is a screenshot from the ASTAR indexing software:
index6
index7

Clearly it is the phi1 angle that is optimized on indexing through the polar transform - the graph then shows the correlation as a function of phi1. Also the template bank has all templates with phi1 = 0.

TL:DR:
I'm quite confident it is phi1 that should be zero, however I think I'm off by a shift of pi/2 on phi2 for some reason.

din14970:
As a small note, my function adjusted with the pi/2 term is also able to reproduce the ASTAR angles in the last image from the previous comment:
image
I think the error is from rounding errors in trying to make the miller indices integers.

pc494:
I think I was being a little slow, because there is a spare degree of freedom we may well be using the same convention and getting "different" answers that both produce the same crystal normal.

If you don't mind, I think the next steps ought to be 

1) check (in python) that the beam directions (cartesian vectors) fill the correct streogram after gridding (fairly sure they do)
2) fix `get_rotation_from_z_to_direction` so that you can take a grid (of vectors) and export them as a grid of euler angles to import correctly to MTEX

The reason for doing it like this is that writing code that makes use of euler angles has always seemed very risky. Especially when we are working in a fairly simple geometry ([1] captures my feelings nicely).

[1] - https://github.com/moble/quaternion/wiki/Euler-angles-are-horrible#opening-issues-and-pull-requests

din14970:
Are you proposing I do this for this PR?

> check (in python) that the beam directions (cartesian vectors) fill the correct streogram after gridding (fairly sure they do)

you mean with some kind of test on the correctness of `get_beam_directions_grid`?

> fix get_rotation_from_z_to_direction so that you can take a grid (of vectors) and export them as a grid of euler angles to import correctly to MTEX

Yes obviously if there's already a function for doing this then my `beam_directions_grid_to_euler` is unnecessary. But correct me if I'm wrong, but it would seem that `get_rotation_from_z_to_direction` gives the "active" rotation from z onto the crystal direction, whereas I've read that crystal orientations are always defined in the "passive" way. So should the function not be `get_rotation_from_direction_to_z`? I'll check what I get in MTEX.

An additional comment: I noticed that you need a crystal structure in `get_rotation_from_z_to_direction` which is obviously the case when "direction" represents a crystal direction [uvw] or (hkl). However, in the case of the grid, one should not interpret these directly as crystal vectors. Variation among lattice parameters will mean that you significantly distort the grid that way. For example, if you have a tetragonal material with say a=1 and c=10, then the number of grid points between [001] and [101] will be the same as the number of gridpoints between [100] and [101], but the angular deviation between the latter will be much greater. So the grid of vectors should be converted to euler angles without taking into consideration the underlying lattice.

Perhaps this calls for having a separate grid conversion function?

pc494:
> Are you proposing I do this for this PR?

I can do it. I will just plot the streogram directly using the formulae from wikipedia.

> > fix get_rotation_from_z_to_direction so that you can take a grid (of vectors) and export them as a grid of euler angles to import correctly to MTEX: 

> Yes obviously if there's already a function for doing this then my `beam_directions_grid_to_euler` is unnecessary. 

Change of heart. I think I got up to speed on what you're trying to do (bigger picture). I had always expected to do template matching in two phases. Radial. Get results. In plane. But you are planning on doing it in a single swoop, which means you're euler angles need to do "good" things (like be 0 at an inplane direction or whatever) - which your function does (or would do if it's broken?!) and the current one doesn't.

>But correct me if I'm wrong, but it would seem that `get_rotation_from_z_to_direction` gives the "active" rotation from z onto the crystal direction, whereas I've read that crystal orientations are always defined in the "passive" way. So should the function not be `get_rotation_from_direction_to_z`? I'll check what I get in MTEX.

I think there are two separate things. "passive" and "active" rotations tell us whether our axes travel with the rotation of the object (all our code is "passive", the axes move with the system), see for example section 2.3 of [1]. What you seem to be talking about is if we define rotation as "crystal2lab" or "lab2crystal". This has been brought to my attention before, very broadly SPED leans one way and EBSD the other & the two are related by an inversion.


> An additional comment: I noticed that you need a crystal structure in `get_rotation_from_z_to_direction` which is obviously the case when "direction" represents a crystal direction [uvw] or (hkl). However, in the case of the grid, one should not interpret these directly as crystal vectors. Variation among lattice parameters will mean that you significantly distort the grid that way. For example, if you have a tetragonal material with say a=1 and c=10, then the number of grid points between [001] and [101] will be the same as the number of gridpoints between [100] and [101], but the angular deviation between the latter will be much greater. So the grid of vectors should be converted to euler angles without taking into consideration the underlying lattice.
> 

Yeah, I had noticed this too, I had been planning on adding an extra keyword argument "cubic" to treat the input as cartesian (which they are here). I think I will still do that but...

> Perhaps this calls for having a separate grid conversion function?

...I also agree with this. 

[1] - https://iopscience-iop-org.ezp.lib.cam.ac.uk/article/10.1088/0965-0393/23/8/083501/meta
[1b] if the links doesn't work, it's meant to point to "Consistent representations of and conversions between 3D rotations" by Rowenhorst et al. (2015)

pc494:
Checked the triangle, looks fine to me, merging.

```
g = get_beam_directions_grid("cubic",2)
from matplotlib import pyplot as plt

X = np.divide(g[:,0],1-g[:,2])
Y = np.divide(g[:,1],1-g[:,2])

plt.scatter(X,Y)
```
conversion formulae from:
https://en.wikipedia.org/wiki/Stereographic_projection

pc494:
This is an excellent addition, many thanks @din14970

Issue #124, opened 2020-09-20
Why are the template libraries so huge? Best way to sample orientations?


din14970:
More of a discussion question.

I'm trying to use pyxem/diffsims/orix for orientation mapping with template matching mainly as an alternative to Astar. However, I've found that the indexing of the same dataset is one to three orders of magnitude slower. This got me digging a bit into the generation of the template libraries.

If I generate a library from a rotation list `rotation_list_generators.get_fundamental_zone_grid(225, 2)`, I need to simulate about 350k templates  and when I index a map of 200x200 with this library this takes many many hours - I suppose because at each pixel it is comparing the DP to each template? So I've resorted to decreasing the sampling substantially in generating the template library. But if I choose something like `rotation_list_generators.get_fundamental_zone_grid(225, 10)`, the IPF-Z plot of the rotations looks really sparse, and I still have more than 2000 templates.

What I wonder is how is Astar able to do this so much faster and still with decent results. Maybe part of the answer is the slowness of python, but I think there's more to it. When I make a library in Astar using the default settings, I get a library of about 1300 templates. Yet they still claim an angular resolution of 1 degree. Indexation of a single pixel is nearly instantaneous and I get an IPF view of the match index with all the other templates. 

Is diffsims sampling orientation space in a really inefficient way? Or is Astar doing things completely differently? 1300 seems like just points to make a fine grid on the 001 - 101 - 111 triangle, but not enough to cover all possible rotations around these zones. My (very non-scientific) feeling is that Astar is just creating a grid of orientations inside the triangle, simulating the DP's on this grid, and rotating the templates to match the experimental DP's during the indexation process. The additional rotation of the template can then be taken into consideration into the indexation. If there is a fast way to rotate the template into agreement with the DP, maybe with some kind of transform similar to a Hough transform, then I guess the library can be generated much more quickly and indexation could be faster too?

pc494:
Good question, I'll try and keep my answer to a sensible length. Before then, for the avoidance of doubt, I must state that Index (from Nanomegas) is propriety closed source code, and the discussions that follow about "what it does" are merely speculation. 

The gridding that we use from orix is consistent with those used in EMSOFT and MTEX (primarily EBSD software) [1], with timings and sizes (for dictionary based indexing) reflective of those seen in the open literature [2].

[...]

PR #137, opened 2020-11-15
Adding support for beam precession

din14970:
What I'm working on here is very premature, opening this PR to get some opinions/feedback and to give you a preview of the code. Apologies for wall of text.

- I fixed a small bug in the `beam_directions_grid_to_euler` function.
- I added the possibility to include the effect of beam precession in the diffraction patterns
- I added/modified some shape factor models
- I modified the way diffraction libraries are generated

**More details**
In the context of my issue #124 and the other discussions we've had elsewhere, I've started looking into the entire chain of operations I would need for template matching with Pyxem. It seems @pc494 you are also kind of working on something in this direction judging by https://github.com/pyxem/pyxem/issues/656?

The things I need for this are:
- a good orientation grid generator (solved in #130)
- simulation of those orientations (currently looking at this point)
- a modified indexer that calculates in-plane rotation, likely via polar transform
- a way to visualize and plot the results; especially quality maps are of interest. I've thrown together a quick stereographic plot functionality for orix but it's too crappy for a PR

The simulated diffraction patterns generated with diffsims looked very odd for my material, in some cases only one diffraction spot was in the pattern. So I looked into:
- the relrod functions in `shape_factor_models`
- the way spot patterns are generated and the available kwargs in `DiffractionGenerator.calculate_ed_data`

With regards to the shape factor function, I saw the default shape is linear. I looked into W&C and at least for a 2-beam case, Howie-Whelan predicts a `(sin(x)/x)^2` - like function, but the exact shape depends also on the extinction distance and thickness. I messed around a bit with the shape factor functions, which you can see here:

afbeelding

`sinc` was a function that was already there, and seems to follow Howie-Whelan quite well. I also tried `sinc^2` but this tapers off the tails too strongly. To get a smooth function that approximately follows the shape of the Howie-Whelan rel-rod I also implemented an `arctan(x)/x` function. I changed all the functions so that the maximum at s=0 hits 1. I figured it doesn't hurt to have a couple more shape factor functions.

To the second point, I felt like the options available for generating a diffraction library "the official way" are a bit lacking:
```
diff_gen = DiffractionGenerator(accelerating_voltage=200, debye_waller_factors)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(library_austenite,
                                           calibration=diffraction_calibration,
                                           reciprocal_radius=reciprocal_radius,
                                           half_shape=(half_size, half_size),
                                           with_direct_beam=False)
```
Many of the options in `calculate_ed_data` don't seem to be available this way. I saw that one used to be able to set `max_excitation_error` in `DiffractionGenerator.__init__` but no longer, what is the reason for this?

Personally, I would like to have full control over the way the patterns are simulated. So I'm inclined to change everything either in the init of `DiffractionGenerator` or in the `get_diffraction_library` function. 

In the process of looking into these diffraction patterns, I added the option to include the effect of beam precession. To calculate this I only use geometry; I don't make any additional assumptions. The Ewald sphere is precessed through the rel rods and the shape factor of each spot is calculated by averaging the shape factor function via integration. The difference is minor but there is some:
afbeelding
Especially further away reflections become a bit brighter with precession. The influence depends on the chosen shape factor model - above I use the `atanc` function. Below is with `linear`
afbeelding
I think the math on the precession is sound but a second pair of eyes might verify this. Of course with this implementation, precession is much more expensive, since I do an integration for each relevant diffraction spot.

The main thing I'm unsure about is which arguments should go where, and why certain arguments were removed. 

**Work in progress?**
- [x] I haven't included any testing yet
- [x] My alteration of the existing functions could break existing tests

#### For reviewers

  later.

### DIFF
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -33,22 +33,22 @@ from diffsims.utils.sim_utils import (
 )
 from diffsims.utils.fourier_transform import from_recip
 from diffsims.utils.shape_factor_models import (
-        linear,
-        atanc,
-        lorentzian,
-        sinc,
-        sin2c,
-        lorentzian_precession,
-        )
+    linear,
+    atanc,
+    lorentzian,
+    sinc,
+    sin2c,
+    lorentzian_precession,
+)
 
 
 _shape_factor_model_mapping = {
-        "linear": linear,
-        "atanc": atanc,
-        "sinc": sinc,
-        "sin2c": sin2c,
-        "lorentzian": lorentzian,
-        }
+    "linear": linear,
+    "atanc": atanc,
+    "sinc": sinc,
+    "sin2c": sin2c,
+    "lorentzian": lorentzian,
+}
 
 
 def _z_sphere_precession(phi, r_spot, wavelength, theta):
@@ -75,15 +75,17 @@ def _z_sphere_precession(phi, r_spot, wavelength, theta):
         The height of the ewald sphere at the point r in A^-1
     """
     phi = np.deg2rad(phi)
-    r = 1/wavelength
+    r = 1 / wavelength
     theta = np.deg2rad(theta)
-    return (-np.sqrt(r**2*(1-np.sin(theta)**2*np.sin(phi)**2) -
-                     (r_spot - r*np.sin(theta)*np.cos(phi))**2) +
-            r*np.cos(theta))
+    return -np.sqrt(
+        r ** 2 * (1 - np.sin(theta) ** 2 * np.sin(phi) ** 2)
+        - (r_spot - r * np.sin(theta) * np.cos(phi)) ** 2
+    ) + r * np.cos(theta)
 
 
-def _shape_factor_precession(z_spot, r_spot, wavelength, precession_angle,
-                             function, max_excitation, **kwargs):
+def _shape_factor_precession(
+    z_spot, r_spot, wavelength, precession_angle, function, max_excitation, **kwargs
+):
     """
     The rel-rod shape factors for reflections taking into account
     precession
@@ -120,29 +122,30 @@ def _shape_factor_precession(z_spot, r_spot, wavelength, precession_angle,
     shf = np.zeros(z_spot.shape)
     # loop over all spots
     for i, (z_spot_i, r_spot_i) in enumerate(zip(z_spot, r_spot)):
+
         def integrand(phi):
-            z_sph = _z_sphere_precession(phi, r_spot_i,
-                                         wavelength, precession_angle)
-            return function(z_spot_i-z_sph, max_excitation, **kwargs)
+            z_sph = _z_sphere_precession(phi, r_spot_i, wavelength, precession_angle)
+            return function(z_spot_i - z_sph, max_excitation, **kwargs)
+
         # average factor integrated over the full revolution of the beam
-        shf[i] = (1/(360))*quad(integrand, 0, 360)[0]
+        shf[i] = (1 / (360)) * quad(integrand, 0, 360)[0]
     return shf
 
 
-def _average_excitation_error_precession(z_spot, r_spot,
-                                         wavelength, precession_angle):
+def _average_excitation_error_precession(z_spot, r_spot, wavelength, precession_angle):
     """
     Calculate the average excitation error for spots
     """
     ext = np.zeros(z_spot.shape)
     # loop over all spots
     for i, (z_spot_i, r_spot_i) in enumerate(zip(z_spot, r_spot)):
+
         def integrand(phi):
-            z_sph = _z_sphere_precession(phi, r_spot_i,
-                                         wavelength, precession_angle)
-            return z_spot_i-z_sph
+            z_sph = _z_sphere_precession(phi, r_spot_i, wavelength, precession_angle)
+            return z_spot_i - z_sph
+
         # average factor integrated over the full revolution of the beam
-        ext[i] = (1/(360))*quad(integrand, 0, 360)[0]
+        ext[i] = (1 / (360)) * quad(integrand, 0, 360)[0]
     return ext
 
 
@@ -172,7 +175,7 @@ class DiffractionGenerator(object):
         Angle about which the beam is precessed. Default is no precession.
     approximate_precession : boolean
         When using precession, whether to precisely calculate average
-        excitation errors and intensities or use an approximation.    
+        excitation errors and intensities or use an approximation.
     shape_factor_model : function or string
         A function that takes excitation_error and
         `max_excitation_error` (and potentially kwargs) and returns
@@ -206,7 +209,9 @@ class DiffractionGenerator(object):
         self.approximate_precession = approximate_precession
         if isinstance(shape_factor_model, str):
             if shape_factor_model in _shape_factor_model_mapping.keys():
-                self.shape_factor_model = _shape_factor_model_mapping[shape_factor_model]
+                self.shape_factor_model = _shape_factor_model_mapping[
+                    shape_factor_model
+                ]
             else:
                 raise NotImplementedError(
                     f"{shape_factor_model} is not a recognized shape factor "
@@ -293,11 +298,11 @@ class DiffractionGenerator(object):
             # We find the average excitation error - this step can be
             # quite expensive
             excitation_error = _average_excitation_error_precession(
-                    z_spot,
-                    r_spot,
-                    wavelength,
-                    self.precession_angle,
-                    )
+                z_spot,
+                r_spot,
+                wavelength,
+                self.precession_angle,
+            )
         else:
             z_sphere = -np.sqrt(r_sphere ** 2 - r_spot ** 2) + r_sphere
             excitation_error = z_sphere - z_spot
@@ -311,25 +316,24 @@ class DiffractionGenerator(object):
         # take into consideration rel-rods
         if self.precession_angle > 0 and not self.approximate_precession:
             shape_factor = _shape_factor_precession(
-                    intersection_coordinates[:, 2],
-                    r_spot,
-                    wavelength,
-                    self.precession_angle,
-                    self.shape_factor_model,
-                    max_excitation_error,
-                    **self.shape_factor_kwargs,
-                    )
+                intersection_coordinates[:, 2],
+                r_spot,
+                wavelength,
+                self.precession_angle,
+                self.shape_factor_model,
+                max_excitation_error,
+                **self.shape_factor_kwargs,
+            )
         elif self.precession_angle > 0 and self.approximate_precession:
             shape_factor = lorentzian_precession(
-                        excitation_error,
-                        max_excitation_error,
-                        r_spot,
-                        self.precession_angle,
-                    )
+                excitation_error,
+                max_excitation_error,
+                r_spot,
+                self.precession_angle,
+            )
         else:
             shape_factor = self.shape_factor_model(
-                excitation_error, max_excitation_error,
-                **self.shape_factor_kwargs
+                excitation_error, max_excitation_error, **self.shape_factor_kwargs
             )
 
         # Calculate diffracted intensities based on a kinematical model.
@@ -356,8 +360,11 @@ class DiffractionGenerator(object):
         )
 
     def calculate_profile_data(
-        self, structure, reciprocal_radius=1.0, minimum_intensity=1e-3,
-        debye_waller_factors={}
+        self,
+        structure,
+        reciprocal_radius=1.0,
+        minimum_intensity=1e-3,
+        debye_waller_factors={},
     ):
         """Calculates a one dimensional diffraction profile for a
         structure.
@@ -469,7 +476,7 @@ class AtomicDiffractionGenerator:
         dtype="float64",
         ZERO=1e-14,
         mode="kinematic",
-        **kwargs
+        **kwargs,
     ):
         """
         Calculates single electron diffraction image for particular atomic
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -357,7 +357,8 @@ class DiffractionGenerator(object):
         )
 
     def calculate_profile_data(
-        self, structure, reciprocal_radius=1.0, minimum_intensity=1e-3
+        self, structure, reciprocal_radius=1.0, minimum_intensity=1e-3,
+        debye_waller_factors={}
     ):
         """Calculates a one dimensional diffraction profile for a
         structure.
@@ -372,6 +373,8 @@ class DiffractionGenerator(object):
         minimum_intensity : float
             The minimum intensity required for a diffraction peak to be
             considered real. Deals with numerical precision issues.
+        debye_waller_factors : dict of str:value pairs
+            Maps element names to their temperature-dependent Debye-Waller factors.
 
         Returns
         -------
@@ -399,7 +402,7 @@ class DiffractionGenerator(object):
             np.asarray(g_hkls),
             prefactor=multiplicities,
             scattering_params=self.scattering_params,
-            debye_waller_factors=self.debye_waller_factors,
+            debye_waller_factors=debye_waller_factors,
         )
 
         if is_lattice_hexagonal(latt):
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -177,7 +177,7 @@ class DiffractionGenerator(object):
         A function that takes excitation_error and
         `max_excitation_error` (and potentially kwargs) and returns
         an intensity scaling factor. If None defaults to
-        `shape_factor_models.lorentzian`. A number of pre-programmed functions
+        `shape_factor_models.linear`. A number of pre-programmed functions
         are available via strings.
     kwargs
         Keyword arguments passed to `shape_factor_model`.
@@ -195,18 +195,14 @@ class DiffractionGenerator(object):
         self,
         accelerating_voltage,
         scattering_params="lobato",
-        precession_angle=None,
-        shape_factor_model=None,
+        precession_angle=0,
+        shape_factor_model="linear",
         approximate_precession=True,
         minimum_intensity=1e-20,
         **kwargs,
     ):
         self.wavelength = get_electron_wavelength(accelerating_voltage)
         self.precession_angle = precession_angle
-        if self.precession_angle is None:
-            self.precession_angle = 0
-        if self.precession_angle < 0:
-            raise ValueError("The precession angle must be >= 0")
         self.approximate_precession = approximate_precession
         if isinstance(shape_factor_model, str):
             if shape_factor_model in _shape_factor_model_mapping.keys():
@@ -217,14 +213,8 @@ class DiffractionGenerator(object):
                     f"model, choose from: {_shape_factor_model_mapping.keys()} "
                     f"or provide your own function."
                 )
-        elif callable(shape_factor_model):
-            self.shape_factor_model = shape_factor_model
-        elif shape_factor_model is None:
-            self.shape_factor_model = lorentzian
         else:
-            raise TypeError(
-                "shape_factor_model must be a recognized string "
-                "or a callable object.")
+            self.shape_factor_model = shape_factor_model
         self.minimum_intensity = minimum_intensity
         self.shape_factor_kwargs = kwargs
         if scattering_params in ["lobato", "xtables"]:
@@ -299,15 +289,16 @@ class DiffractionGenerator(object):
         r_spot = np.sqrt(np.sum(np.square(cartesian_coordinates[:, :2]), axis=1))
         z_spot = cartesian_coordinates[:, 2]
 
-        if self.precession_angle > 0 and not self.approximate_precession:
-            # We find the average excitation error - this step can be
-            # quite expensive
-            excitation_error = _average_excitation_error_precession(
-                    z_spot,
-                    r_spot,
-                    wavelength,
-                    self.precession_angle,
-                    )
+        if self.precession_angle > 0:
+            if not self.approximate_precession:
+                # We find the average excitation error - this step can be
+                # quite expensive
+                excitation_error = _average_excitation_error_precession(
+                        z_spot,
+                        r_spot,
+                        wavelength,
+                        self.precession_angle,
+                        )
         else:
             z_sphere = -np.sqrt(r_sphere ** 2 - r_spot ** 2) + r_sphere
             excitation_error = z_sphere - z_spot
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -208,7 +208,7 @@ class DiffractionGenerator(object):
         accelerating_voltage,
         scattering_params="lobato",
         precession_angle=0,
-        shape_factor_model="linear",
+        shape_factor_model="lorentzian",
         approximate_precession=True,
         minimum_intensity=1e-20,
         **kwargs,
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -289,16 +289,15 @@ class DiffractionGenerator(object):
         r_spot = np.sqrt(np.sum(np.square(cartesian_coordinates[:, :2]), axis=1))
         z_spot = cartesian_coordinates[:, 2]
 
-        if self.precession_angle > 0:
-            if not self.approximate_precession:
-                # We find the average excitation error - this step can be
-                # quite expensive
-                excitation_error = _average_excitation_error_precession(
-                        z_spot,
-                        r_spot,
-                        wavelength,
-                        self.precession_angle,
-                        )
+        if self.precession_angle > 0 and not self.approximate_precession:
+            # We find the average excitation error - this step can be
+            # quite expensive
+            excitation_error = _average_excitation_error_precession(
+                    z_spot,
+                    r_spot,
+                    wavelength,
+                    self.precession_angle,
+                    )
         else:
             z_sphere = -np.sqrt(r_sphere ** 2 - r_spot ** 2) + r_sphere
             excitation_error = z_sphere - z_spot
@@ -317,7 +316,7 @@ class DiffractionGenerator(object):
                     wavelength,
                     self.precession_angle,
                     self.shape_factor_model,
-                    self.max_excitation_error,
+                    max_excitation_error,
                     **self.shape_factor_kwargs,
                     )
         elif self.precession_angle > 0 and self.approximate_precession:
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -32,7 +32,23 @@ from diffsims.utils.sim_utils import (
     get_intensities_params,
 )
 from diffsims.utils.fourier_transform import from_recip
-from diffsims.utils.shape_factor_models import linear
+from diffsims.utils.shape_factor_models import (
+        linear,
+        atanc,
+        lorentzian,
+        sinc,
+        sin2c,
+        lorentzian_precession,
+        )
+
+
+_shape_factor_model_mapping = {
+        "linear": linear,
+        "atanc": atanc,
+        "sinc": sinc,
+        "sin2c": sin2c,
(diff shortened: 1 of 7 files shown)


pc494:
A little difficult to structure a response, as there is a lot of very good stuff here. So I've just worked through it point by point, I think that will get us going in the right direction.

> **What does this PR do? Please describe and/or link to an open issue.**
> 
>     * I fixed a small bug in the `beam_directions_grid_to_euler` function.
> 
>     * I added the possibility to include the effect of beam precession in the diffraction patterns
> 
>     * I added/modified some shape factor models
> 
>     * I modified the way diffraction libraries are generated

The first three points look good, I'll have a look at the changes for diffraction library generation once I've finished this comment

> 
> **More details**
> In the context of my issue #124 and the other discussions we've had elsewhere, I've started looking into the entire chain of operations I would need template matching with Pyxem. It seems @pc494 you are also kind of working on something in this direction judging by [pyxem/pyxem#656](https://github.com/pyxem/pyxem/issues/656)?

I'm not really working on this at the moment, but once we had your gridding in diffsims, it seemed worth making it clear in pyxem that the workflow you describe is an option (and probably a target)

> simulation of those orientations (currently looking at this point)

So I think you can create those simulations with the code as it currently is, but it's clear that we want to improve the pipeline in this regard

>  a modified indexer that calculates in-plane rotation, likely via polar transform

I have thoughts about this, but I think I'll leave them to somewhere else
 
> a way to visualize and plot the results; especially quality maps are of interest. I've thrown together a quick stereographic plot functionality for orix but it's too crappy for a PR

This one is a mixture of pyxem and orix (CrystalMap) and should be fairly simple to deal with. https://github.com/pyxem/pyxem/pull/654 will be of interest.



> The simulated diffraction patterns generated with diffsims looked very odd for my material, in some cases only one diffraction spot was in the pattern. 

In my experience this tends to happens when your relrods are too short.

> So I looked into:
> 
> the relrod functions in `shape_factor_models`

Think this is probably a case of the more the merrier


>  the way spot patterns are generated and the available kwargs in `DiffractionGenerator.calculate_ed_data`

> To the second point, I felt like the options available for generating a diffraction library "the official way" are a bit lacking:
> 
> ```
> diff_gen = DiffractionGenerator(accelerating_voltage=200, debye_waller_factors)
> lib_gen = DiffractionLibraryGenerator(diff_gen)
> diff_lib = lib_gen.get_diffraction_library(library_austenite,
>                                            calibration=diffraction_calibration,
>                                            reciprocal_radius=reciprocal_radius,
>                                            half_shape=(half_size, half_size),
>                                            with_direct_beam=False)
> ```

This is a bug, `get_diffraction_library()` should have `kwargs` that get passed to `calculate_ed_data`. Good find.

> Many of the options in `calculate_ed_data` don't seem to be available this way. I saw that one used to be able to set `max_excitation_error` in `DiffractionGenerator.__init__` but no longer, what is the reason for this?

Philosophical. The `DiffractionGenerator` is meant to be the _Microscope_, with _sample_ properties specified as arguments to `calculate_ed_data`, so we shuffled `max_excitation_error` around during a recent refactor.

> Personally, I would like to have full control over the way the patterns are simulated. So I'm inclined to change everything either in the init of `DiffractionGenerator` or in the `get_diffraction_library` function.

Does the solution proposed above cover these needs?

> In the process of looking into these diffraction patterns, I added the option to include the effect of beam precession. To calculate this I only use geometry; I don't make any additional assumptions. The Ewald sphere is precessed through the rel rods and the shape factor of each spot is calculated by averaging the shape factor function via integration. The difference is minor but there is some: I think the math on the precession is sound but a second pair of eyes might verify this. Of course with this implementation, precession is much more expensive, since I do an integration for each relevant diffraction spot.

Have you seen https://journals-iucr-org.ezp.lib.cam.ac.uk/b/issues/2019/04/00/je5015/ ? I think that would save you doing the integrations (and could be written in as a "shape factor model").

pc494 (commented):
Had a read of the code, but I think sorting out the "descriptive" stuff before attempting a code review is wise.

din14970:
Thanks for the quick  thorough response, sorry for always hitting you with these long posts. They help me keep the overview myself.

> In my experience this tends to happens when your relrods are too short.

Yes, exactly, which got me looking into how it was working in the background. It then seems that the default `max_excitation_error=1e-2` is quite small, 1e-1 gave me better results. I won't mess with defaults though, I will leave that up to the maintainers.

> This is a bug, get_diffraction_library() should have kwargs that get passed to calculate_ed_data. Good find.
> Philosophical. The DiffractionGenerator is meant to be the Microscope, with sample properties specified as arguments to calculate_ed_data, so we shuffled max_excitation_error around during a recent refactor.

Ok, that clarifies a lot for me, thanks. But if the sample dependent things are to be split of, would the following also not fit best in `get_diffraction_library`?:

*  `debye_waller_factors`
*  `scattering_params`
* `shape_factor_model`
* `kwargs` for `shape_factor_model`

> Have you seen https://journals-iucr-org.ezp.lib.cam.ac.uk/b/issues/2019/04/00/je5015/ ? 

Do you have a DOI link or so to this? I don't seem to have access. Indeed having an approximation for the precession wouldn't be a bad idea, especially for the library generation. But you might as well have the option for a full calculation if you only need a few DP's for some reason.

> Had a read of the code, but I think sorting out the "descriptive" stuff before attempting a code review is wise.

No worries, as I said it's still a preliminary mess, but in this way you can check the progress if you want

pc494:
> Thanks for the quick thorough response, sorry for always hitting you with these long posts. They help me keep the overview myself.
> 
> > In my experience this tends to happens when your relrods are too short.
> 
> Yes, exactly, which got me looking into how it was working in the background. It then seems that the default `max_excitation_error=1e-2` is quite small, 1e-1 gave me better results. I won't mess with defaults though, I will leave that up to the maintainers.

I think for a while we had written in the documentation described as often 1/sample_thickness and 1e-1 corresponded to a very thin sample (10nm maybe, I would have to check)

> > This is a bug, get_diffraction_library() should have kwargs that get passed to calculate_ed_data. Good find.
> > Philosophical. The DiffractionGenerator is meant to be the Microscope, with sample properties specified as arguments to calculate_ed_data, so we shuffled max_excitation_error around during a recent refactor.
> 
> Ok, that clarifies a lot for me, thanks. But if the sample dependent things are to be split of, would the following also not fit best in `get_diffraction_library`?:

> 
>     * `debye_waller_factors`
>     * `scattering_params`

I'm not sure. They are currently quiet underdeveloped parts of the code and I would have to investigate before coming to a conclusion.

>     * `shape_factor_model`
>     * `kwargs` for `shape_factor_model`
> 

I think adding a **kwargs options to `get_diffraction_library` would mean that you could do those at that level? I would expect:

```python 
diff_lib = lib_gen.get_diffraction_library(library_austenite,
                                           calibration=diffraction_calibration,
                                           reciprocal_radius=reciprocal_radius,
                                           half_shape=(half_size, half_size),
                                           with_direct_beam=False,
                                           shape_factor_model="linear")
```
 
to work.

> Have you seen https://journals-iucr-org.ezp.lib.cam.ac.uk/b/issues/2019/04/00/je5015/ ? Do you have a DOI link or so to this? I don't seem to have access.

https://doi.org/10.1107/S2052520619007534

Sorry, forgot that my browser extension makes a mess of copy+pasting links.

> Indeed having an approximation for the precession wouldn't be a bad idea, especially for the library generation. But you might as well have the option for a full calculation if you only need a few DP's for some reason.

Yes, I agree.

din14970:
@pc494 Thanks for the link to the paper, I've implemented both the lorentzian and the closed form approximation for precession using the lorentzian. I modified the function slightly from the paper so that it also takes the parameter `max_excitation_error`, defined as 1/thickness. I've also shuffled around the parameters to align it with the idea `DiffractionGenerator` = instrument state/physics model and `get_diffraction_library`= sample/material parameters. The split I've made (new parameters I've introduced are in bold):

`DiffractionGenerator` stores values:
* accelerating_voltage
* scattering_params
* **minimum_intensity**
* **precession_angle**
* **approximate_precession**
* shape_factor_model
* kwargs for shape factor model

`get_diffraction_library` accepts arguments:
* structure_library
* calibration
* reciprocal_radius
* half_shape
* with_direct_beam
* max_excitation_error
* debye_waller_factors

I've decided that shape_factor_model is rather something related to the model we choose, rather than something dependent on the material. max_excitation_error and debye_waller_factors do depend on the sample.

To me this is a logical split, and I think it is now more or less ready for my purposes. All the docstring documentation should be ok. Before I start messing around in the tests, I wanted to hear your thoughts on this way of organizing the parameters. If ok, I will update the broken tests and write new tests for the stuff I added.

pc494:
@din14970, at a first glance this look like a good set of decisions to me. I'll try to get a review of the code back to you later today or tomorrow if that's okay with you.

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +164,69 @@ class DiffractionGenerator(object):
...
+    approximate_precession=False the custom shape_factor_model is used.
     """
 
     def __init__(
         self,
         accelerating_voltage,
-        max_excitation_error=None,
-        debye_waller_factors={},
         scattering_params="lobato",
+        precession_angle=None,
```
pc494: would default this to `0`

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +164,69 @@ class DiffractionGenerator(object):
...
+        shape_factor_model=None,
```
pc494: I think this is clearer if you write explicitly that it defaults to `"linear"`

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +164,69 @@ class DiffractionGenerator(object):
...
-            print(
-                "This class changed in v0.3 and no longer takes a maximum_excitation_error"
-            )
         self.wavelength = get_electron_wavelength(accelerating_voltage)
-        self.debye_waller_factors = debye_waller_factors
+        self.precession_angle = precession_angle
+        if self.precession_angle is None:
+            self.precession_angle = 0
+        if self.precession_angle < 0:
+            raise ValueError("The precession angle must be >= 0")
```
pc494: Don't think we need to check this case, just let it run to error, users shouldn't be putting negative numbers in here.

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +164,69 @@ class DiffractionGenerator(object):
...
+        if isinstance(shape_factor_model, str):
+            if shape_factor_model in _shape_factor_model_mapping.keys():
+                self.shape_factor_model = _shape_factor_model_mapping[shape_factor_model]
+            else:
+                raise NotImplementedError(
+                    f"{shape_factor_model} is not a recognized shape factor "
+                    f"model, choose from: {_shape_factor_model_mapping.keys()} "
+                    f"or provide your own function."
+                )
+        elif callable(shape_factor_model):
```
pc494: would make this an `else` and drop the `is None` clause

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +164,69 @@ class DiffractionGenerator(object):
...
+            self.shape_factor_model = shape_factor_model
+        elif shape_factor_model is None:
+            self.shape_factor_model = lorentzian
+        else:
```
pc494, line 230: again, don't think we should type check here.

diffsims/generators/diffraction_generator.py
```diff
@@ -149,21 +297,50 @@ def calculate_ed_data(
...
+        r_spot = r_spot[intersection]
+        g_indices = spot_indices[intersection]
         g_hkls = spot_distances[intersection]
-
-        if shape_factor_model is not None:
-            shape_factor = shape_factor_model(
-                excitation_error, max_excitation_error, **kwargs
-            )
+        # take into consideration rel-rods
+        if self.precession_angle > 0 and not self.approximate_precession:
```
pc494, line 326: picky, but I would probably find these easier to read if they were nested:
```python
if self.precession_angle > 0:
        if not self.approximate_precession: 
        
        else:
else:
```

diffsims/generators/sphere_mesh_generators.py
```diff
@@ -481,5 +481,5 @@ def beam_directions_grid_to_euler(vectors):
     phi2 = sign * np.nan_to_num(np.arccos(x_comp / norm_proj))
     # phi1 is just 0, rotation around z''
     phi1 = np.zeros(phi2.shape[0])
-    grid = np.rad2deg(np.vstack([phi1, Phi, phi2]).T)
+    grid = np.rad2deg(np.vstack([phi1, Phi, np.pi/2 - phi2]).T)
```
pc494: This was the bugfix? Could you remind me what's going on?

diffsims/utils/shape_factor_models.py
```diff
@@ -77,7 +77,122 @@ def sinc(excitation_error, max_excitation_error, minima_number=5):
...
+    num = np.sin(fac * excitation_error)
+    denom = fac * excitation_error
+    return np.nan_to_num(
+            np.abs(np.divide(num, denom, out=np.zeros_like(num), where=denom != 0)),
+            nan=1,
+            )
 
-    num = np.sin(np.pi * minima_number * excitation_error / max_excitation_error)
-    denom = excitation_error
-    return np.abs(np.divide(num, denom, out=np.zeros_like(num), where=denom != 0))
```
pc494: I'm not sure these are quiet the same?

diffsims/utils/shape_factor_models.py
```diff
@@ -77,7 +77,122 @@ def sinc(excitation_error, max_excitation_error, minima_number=5):
...
+        The minima_number'th minima lies at max_excitation_error from 0
+
+    Returns
+    -------
+    intensity : array-like or float
+    """
+    return sinc(excitation_error, max_excitation_error, minima_number)**2
+
+
+def atanc(excitation_error, max_excitation_error):
```
pc494: this should have a `minima_number` I think?

diffsims/utils/shape_factor_models.py
```diff
@@ -77,7 +77,122 @@ def sinc(excitation_error, max_excitation_error, minima_number=5):
...
+
+    max_excitation_error : float
+        The distance at which a reflection becomes extinct
+
+    Returns
+    -------
+    intensity_factor : array-like or float
+        Vector representing the rel-rod factor for each reflection
+
+    Notes
```
pc494: References is preferred here.

diffsims/utils/shape_factor_models.py
```diff
@@ -77,7 +77,122 @@ def sinc(excitation_error, max_excitation_error, minima_number=5):
...
+    intensity : array-like or float
+    """
+    fac = np.pi * 5 / np.abs(max_excitation_error)
+    return np.nan_to_num(
+            np.arctan(fac*excitation_error)/(fac*excitation_error),
+            nan=1,
+            )
+
+
+def lorentzian(excitation_error, max_excitation_error):
```
pc494, line 139: I (@pc494) will double check these implementations at the end of this PR.

diffsims/generators/diffraction_generator.py
```diff
@@ -31,7 +32,118 @@
...
+
+_shape_factor_model_mapping = {
+        "linear": linear,
+        "atanc": atanc,
+        "sinc": sinc,
+        "sin2c": sin2c,
+        "lorentzian": lorentzian,
+        }
+
+
```
pc494, line 53: Will review these at the end of the PR as these are a bit more involved

pc494 (changes requested):
This looks like a very substantial improvement. Once it's got some tests I'll have another pass and try my best to double check all the physics.

diffsims/utils/shape_factor_models.py
din14970: to be consistent with the sinc functions yes, but its a bit awkward since it's a smooth function. And just like max_excitation error, the parameter just rescales in the x-direction. I added it, it can be removed later if necessary. In any case, I would expect the Lorentzian to become the default, since there is a reference for it.

diffsims/utils/shape_factor_models.py
din14970: no they are not. The denominator is scaled by the factor `fac` so that the maximum at s=0 comes out to 1. In my opinion that's a bit nicer to plot and compare for example to the linear case. In any case it shouldn't matter since the intensities get rescaled to the maximum at a later stage, and each reflection is multiplied by this same factor, right?

diffsims/generators/sphere_mesh_generators.py
din14970: Exactly. This is to match the triangle of beam directions to the ones defined. I found there was a 90 degree offset in the original implementation. Without the fix, the triangle for cubic looks like:
image
This is the triangle 001 - 111 - 011.

With the fix it looks like:
image
So the triangle is between (001) - (111) - (101) as defined in `rotation_list_generators`

I checked also for the other crystal systems whether it's correct

diffsims/generators/diffraction_generator.py
din14970: There is no nested else though. Only if one has precession AND one wants a full calculation does the mean excitation error need to be calculated. In all other cases we just calculate it the simple way. I'll implement it as you say and one can return to it later.

diffsims/generators/diffraction_generator.py
din14970: I'll leave it at linear, but since there is a reference to back up Lorenzian, I would vote to change this to be the default.

diffsims/generators/sphere_mesh_generators.py
pc494: Clear improvement, good to see.

diffsims/utils/shape_factor_models.py
pc494: Makes perfect sense, it looked like a refactor (from the code) but I had forgot you said you were scaling everything to have unit intensity, in which case this is fine and good.

diffsims/generators/diffraction_generator.py
pc494: I thought one could refactor the approximation inside?
```python
if self.precession_angle > 0:
        if not self.approximate_precession: 
        
        elif self.approximate_precession:

elif self.precession_angle == 0:
```
(I've made everything `elif` in this comment to aid clarity)

dnjohnstone:
Just to say - this looks great, thanks! I'd quite like to review it though at some point, can someone ping me when appropriate.

diffsims/generators/diffraction_generator.py
din14970: Unfortunately I've changed it back to what it was originally, if you find a better way feel free to edit. Reason is that in the `elif self.approximate_precession` i would have the same code as in the `elif self.precession_angle == 0`, which would be unnecessary duplication. This only influences how the `excitation_error` is calculated and which spots need to be thrown out in the first stage. If there is precession but we are content with the approximation, then we calculate the excitation error in the same way as if there was no precession. If the precession is 0 but we set approximate precession to True, we also want to execute the else block.

din14970:
@dnjohnstone @pc494 I would tentatively say it's ready for a more thorough review. The functionality that I wanted is there. The test coverage is again 100%, but correctness of the physics is of course best verified with additional pairs of eyes. @pc494 I didn't implement 1 of your comments above, hope you understand my reasoning.

pc494:
I will aim to review this tomorrow, and then ping dnjohnstone, it might take a week or so to get this all done, I hope that's okay with you @din14970.

din14970:
@pc494 I'm in no rush, it's your package take your time. I'm working on the next step for the indexation.

dnjohnstone:
> @pc494 I'm in no rush, it's your package take your time. I'm working on the next step for the indexation.

Just to say @din14970 - I think we're getting to the point where if you're planning to continue contributing around the level you have been recently, then we can have a conversation about making "your package" more like "our package" - ping us an e-mail at pyxem.team@gmail.com and we can set that chat up... It would certainly be good for us to add another person to the core developer team for diffsims before too long if something like that might fit with your personal roadmap.

pc494:
I made an algebraic error, which means I thought the function below was incorrect, it's not, but for clarity I reproduce my working and thinking, the conclusion of which is that while this could be better documented, it is essentially correct.

```python
def _z_sphere_precession(phi, r_spot, wavelength, theta):
    """
    Returns the z-coordinate in reciprocal space of the Ewald sphere at
    distance r from the origin along the x-axis
    Parameters
    ----------
    phi : float
        The angle the beam is currently precessed to in degrees.
        If the beam were projected on the x-y plane, the angle
        is the angle between this projection with the x-axis.
    r_spot : float
        The distance to the point on the x-axis where we calculate z in A^-1
    wavelength : float
        The electron wavelength in A^-1
    theta : float
        The precession angle in degrees (angle between beam and optical axis)
    Returns
    -------
    z : float
        The height of the ewald sphere at the point r in A^-1
    """
    phi = np.deg2rad(phi)
    r = 1 / wavelength
    theta = np.deg2rad(theta)
    return -np.sqrt(
        r ** 2 * (1 - np.sin(theta) ** 2 * np.sin(phi) ** 2)
        - (r_spot - r * np.sin(theta) * np.cos(phi)) ** 2
    ) + r * np.cos(theta)
```

So that we're clear on our terms.
`theta` is the precession angle, simple enough, in my sketch (of the x-z plane) it's inplane
`phi` is the azimuthal angle, which we are integrating over, it forms a circle about the z axis which runs through (0,0)
We're using the `wavelength` to calculate an Ewald Sphere radius, in what follows I will denote this radius `R`

So we arrive at perhaps my first misunderstanding, `r_spot`. I think it's the two dimensional distance from our spot to the origin, excluding height. So for a spot (x,y,z) it is equal in value to sqrt(x ** 2 + y ** 2). Although this is confusing with the current documentation this is an easy fix, it does actually simplify the algebra a bit, so fair enough.

So, I'm now going to try and reproduce the final formula.

Firstly, our origin is the origin of the reciprocal lattice (0,0,0). For our given geometry the center of the Ewald sphere is a function of `phi` (and `theta` and `R`)

(R sin theta cos phi , R sin theta sin phi R cos theta)  = {for readability later} (A,B,C)

Meanwhile our spot lies at

( X , Y,  Z )

where Z will be our output. Assuming what I wrote about `r_spot` was correct, and because we will always integrate over everything (I don't think this would work if we weren't integrating a full cycle) we can replace our spot with

(r_spot, 0 , Z)

We want to solve for Z, which we do by stating that the distance between these two things will be R

So we get to

R**2 =  (r_spot - A) ** 2 + (Z-C) ** 2 + B ** 2
(Z - C)**2 = (- (r_spot - A) ** 2 - B ** 2 + R ** 2)
Z = C - sqrt((r_spot - A) ** 2 + B ** 2 - R^2)

Compared to:

```python
-np.sqrt(
        r ** 2 * (1 - np.sin(theta) ** 2 * np.sin(phi) ** 2)
        - (r_spot - r * np.sin(theta) * np.cos(phi)) ** 2
    ) + r * np.cos(theta)
```
So the last term in the python code is C (great), and the `1` in the first set of braces is R**2, the term beyond is B and the r_spot - A term is also good.

Excellent, and expensive set of typing for an algebra mistake on my part, but we can now be fairly sure that this function is correct. Going to go do something else for a bit and then will see what's next for review.

diffsims/generators/sphere_mesh_generators.py
```diff
@@ -481,5 +481,5 @@ def beam_directions_grid_to_euler(vectors):
     phi2 = sign * np.nan_to_num(np.arccos(x_comp / norm_proj))
     # phi1 is just 0, rotation around z''
     phi1 = np.zeros(phi2.shape[0])
-    grid = np.rad2deg(np.vstack([phi1, Phi, phi2]).T)
+    grid = np.rad2deg(np.vstack([phi1, Phi, np.pi / 2 - phi2]).T)
```
dnjohnstone, line 484: Where did this one come from? (sorry I've probably just missed this)

dnjohnstone (changes requested):
looking pretty good here - some relatively superficial points.

dnjohnstone:
Also, don't forget a changelog bump

dnjohnstone:
@din14970 @pc494 - this isn't really the right place, but thought you might both like a look at this paper if you haven't seen it yet: https://scripts.iucr.org/cgi-bin/paper?iv5009

diffsims/utils/shape_factor_models.py
```diff
@@ -77,7 +77,126 @@ def sinc(excitation_error, max_excitation_error, minima_number=5):
...
+    sigma = np.pi/max_excitation_error
+    fac = sigma/(np.pi*(sigma**2*excitation_error**2+1))
+    return fac
+
+
+def lorentzian_precession(excitation_error, max_excitation_error,
+                          r_spot, precession_angle):
+    """
+    Intensity profile factor for a precessed beam assuming a Lorentzian
+    intensity profile for the un-precessed beam. After [1].
```
pc494: As above:

`After [1] ---> This is equation (10) in the reference [1].`

pc494 (commented):
The Lorenzian functions look good to me.

pc494:
> @pc494 I'm in no rush, it's your package take your time. I'm working on the next step for the indexation.

Which step have you moved on to?

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +167,61 @@ class DiffractionGenerator(object):
     ----------
     accelerating_voltage : float
         The accelerating voltage of the microscope in kV.
-    max_excitation_error : float
-        Removed in this version, defaults to None
-    debye_waller_factors : dict of str:value pairs
-        Maps element names to their temperature-dependent Debye-Waller factors.
     scattering_params : str
         "lobato" or "xtables"
+    minimum_intensity : float
```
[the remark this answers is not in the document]
din14970: before it was just hard-coded as 1e-20 inside the function, I figured it might as well be a parameter. I guess some absolute value cut-off is most fair across different diffraction patterns in the library, resembling a kind of detection limit of a detector. On the other hand I guess different phases would require different cut-offs. Is this logic correct?

diffsims/generators/diffraction_generator.py
```diff
@@ -52,27 +167,61 @@ class DiffractionGenerator(object):
...
     """
 
     def __init__(
         self,
         accelerating_voltage,
-        max_excitation_error=None,
-        debye_waller_factors={},
         scattering_params="lobato",
+        precession_angle=0,
+        shape_factor_model="linear",
```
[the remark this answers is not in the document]
din14970: My vote is on Lorenzian since we can point to a reference and should approximate a 2-beam rel-rod. Linear seems a bit arbitrary.

diffsims/generators/sphere_mesh_generators.py
din14970: see https://github.com/pyxem/diffsims/pull/137#discussion_r525348812

din14970:
> contributing around the level you have been recently

While I don't mind helping out, I don't know whether I can commit to this for certain at this point. I'm now dealing with a specific problem which I would like solved within reasonable time, and pyxem/diffsims provides the perfect scaffold for this. It is mostly my frustrations with the rigidity of ASTAR that is driving me at this point. Still I'd like to hear from you where you think the package is headed. I'll send you an e-mail.

> Which step have you moved on to?

The indexation which simultaneously solves for in plane angle. I'm close to a kind of prototype which I'll share maybe in a Pyxem issue. Then we can discuss how to best integrate it there.

pc494:
email has arrived in the correct box and we can continue that conversation there.

> > Which step have you moved on to?
> 
> The indexation which simultaneously solves for in plane angle. I'm close to a kind of prototype which I'll share maybe in a Pyxem issue. Then we can discuss how to best integrate it there.

Yep sounds good to me. Fingers crossed, it should be fairly easy to slot into the framework we currently have.

diffsims/generators/diffraction_generator.py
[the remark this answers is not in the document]
pc494: Yes. Let's change this to Lorenzian.

pc494 (approved):
There is one default that we should change, but other than that this is good to go as far as I am concerned.

dnjohnstone (approved):

Issue #35, opened 2019-11-11
Noise models for simulations


dnjohnstone:
Add some functions to add noise to simulations. E.g:

def _addnoise(x, alpha, mu, sigma):
    """Add Poisson-Gaussian noise to the data x

    Parameters
    ----------
    x : float
        The original data

    alpha : float
        Level of noise gain

    mu : float
        Level of noise offset

    sigma : float
        Level of Gaussian noise

    Returns
    -------
    y : float
        The corrupted data

    """
    y = alpha * np.random.poisson(x / alpha) + mu + sigma * np.random.randn()
    return y

def add_poisson_gaussian_noise(X, alpha, mu, sigma):
    """Add Poisson-Gaussian noise to the data X

    Parameters
    ----------
    X : array
        The data to be corrupted

    alpha : float
        Level of noise gain

    mu : float
        Level of noise offset

    sigma : float
        Level of Gaussian noise

    Returns
    -------
    Y : array
        The corrupted data

    """
    # Do some error checking
    if alpha < 0. or alpha > 1.:
        raise ValueError("alpha should be in range [0,1]")
    # Vectorize noise function
    addnoise = np.vectorize(_addnoise, otypes=[np.float])

    # Rescale to [0,1] range
    Xmax = np.amax(X)
    X = X / Xmax

    # Add noise
    Y = addnoise(X, alpha, mu, sigma)

    # Rescale to [0,1] range
    Y = Y + np.abs(np.amin(Y))

    # Rescale to X range
    Y = Xmax * Y / np.amax(Y)
    return Y

    def _gamma_noise_model(x, a, mu):
        """Calculates the expected variance of a pixel based on its value `x`"""
        return a * (x / mu) * np.exp(- x / mu)

    def add_parameterized_gamma_noise(self, a=75., mu=10.):
        """Adds noise based on a custom model to the data."""
        noise = np.random.normal(0., self._gamma_noise_model(self.data, a, mu)**0.5)
        original_dtype = self.data.dtype
        self.data = (
            self.data.astype(noise.dtype) + noise
        ).astype(original_dtype)
        self.events.data_changed.trigger(obj=self)

pc494:
These were brought into version 0.4.0

Issue #141, opened 2020-12-05
diffsims 0.4.0


pc494:
It seems very likely that we are going to want some of the additions to 0.4.0 to integrate into the forthcoming pyxem release. So we should start thinking about packaging what we've got.

Outstanding code:
- [x] Add a crude noise generator
- [x] ReciprocalLatticePoint behaves as expected with only one point #125
- [x] ~ReciprocalLatticePoint.families attribute, which lists the {hkl} families for all (hkl) in the list of points #133~
- [x] Random sphere mesh generator (#142)
- [x] ~Close issue #144~ 
- [x] Rename the testing folders so the test_ prefix is omitted (**FINAL** thing so as to avoid merge conflicts)

Points of note:
I will make the release of diffsims just **before** we release pyxem. This means that we can then set pyxem to depend directly on `0.4.0` from it's latest release onwards.

EDIT: calibration workflow component dropped, noise generator added.

hakonanes:
Sounds good, will look at the mentioned issues this week. I don't have anything else at the moment.

pc494:
This is now targetted for Wednesday 16th.

pc494:
I have dropped closing #88 as a bit too much for now.

hakonanes:
I've addressed #125 in PR #145 and #133 in #146, but I'm not sure if I can get these done so that you can release diffsims tomorrow...

I don't have anything else after these.

pc494:
Yeah, we're going to miss Wednesday 16th, but everything that needs doing is open and in public (well apart from a PR #145, but that's not going to be massive). I am going to package up on Monday 21st if all things go smoothly (possibly as a `rc` if required to keep pyxem 0.12.3 running)

hakonanes:
> (well apart from a PR #145, but that's not going to be massive)

What do you mean by this? As in that one is done, but the other PRs aren't? Because I think the `ReciprocalLatticePoint` class now treats having one vector OK.

pc494:
Sorry, I got the wrong PR number, I meant #144, which is about moving point--->vector

pc494:
Our third release candidate should do the trick, with a full release at the end of the week.

hakonanes:
I've moved remaining issues tagged with v0.4.0 to either v0.4.1 or v0.5.0

PR #143, opened 2020-12-15
Adds the "pattern" module with detector functions

pc494:
---
name: Adds the "pattern" module
about: Addressing the (now top) box on #141 - more details to come.

---

Introduces a patterns "folder" with a single module, `detector_functions` that, broadly, takes a 2D pattern and adds "detector noise"

As an aside, this route was chosen as potential future modules like `probe_functions` will slot nicely into this framework

#### For reviewers

  later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - API reference documentation via Read The Docs: https://diffsims.rtfd.io
 - New module: `sphere_mesh_generators`
 - New module: `detector_functions`
+- New module: `ring_pattern_utils`
 - beam precession is now supported in simulating electron diffraction patterns
 - more shape factor functions have been added
 - This project now keeps a Changelog
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - API reference documentation via Read The Docs: https://diffsims.rtfd.io
 - New module: `sphere_mesh_generators`
+- New module: `detector_fucntions`
 - beam precession is now supported in simulating electron diffraction patterns
 - more shape factor functions have been added
 - This project now keeps a Changelog
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - API reference documentation via Read The Docs: https://diffsims.rtfd.io
 - New module: `sphere_mesh_generators`
-- New module: `detector_fucntions`
+- New module: `detector_functions`
 - beam precession is now supported in simulating electron diffraction patterns
 - more shape factor functions have been added
 - This project now keeps a Changelog
--- a/diffsims/pattern/detector_functions.py
+++ b/diffsims/pattern/detector_functions.py
@@ -19,6 +19,7 @@
 from numpy.random import default_rng
 from scipy import ndimage as ndi
 
+
 def _process_seed_argument(seed):
     """ Sets up a numpy random number generator with a seed"""
     if seed is not None:
@@ -28,8 +29,9 @@ def _process_seed_argument(seed):
 
     return rng
 
-def constrain_to_dynamic_range(pattern,detector_max=None):
-    """ Force the values within pattern to lie between [0,detector_max]
+
+def constrain_to_dynamic_range(pattern, detector_max=None):
+    """Force the values within pattern to lie between [0,detector_max]
 
     Parameters
     ----------
@@ -49,6 +51,7 @@ def constrain_to_dynamic_range(pattern,detector_max=None):
         pattern[pattern > detector_max] = detector_max
 
     return pattern
+
 
 def add_gaussian_blur(pattern, sigma):
     """
@@ -94,7 +97,8 @@ def add_shot_noise(pattern, seed=None):
 
     return rng.poisson(pattern)
 
-def add_gaussian_noise(pattern,sigma,seed=None):
+
+def add_gaussian_noise(pattern, sigma, seed=None):
     """
     Applies gaussian noise at each pixel within the pattern
 
@@ -112,13 +116,13 @@ def add_gaussian_noise(pattern,sigma,seed=None):
     corrupted_pattern :
     """
     rng = _process_seed_argument(seed)
-    pertubations = rng.normal(loc=0,scale=sigma,shape=pattern.shape)
+    pertubations = rng.normal(loc=0, scale=sigma, shape=pattern.shape)
     pattern = pattern + pertubations
 
     return constrain_to_dynamic_range(pattern)
 
 
-def add_dead_pixels(pattern,n=None,fraction=None,seed=None):
+def add_dead_pixels(pattern, n=None, fraction=None, seed=None):
     """
     Adds randomly placed dead pixels onto a pattern
 
@@ -141,7 +145,7 @@ def add_dead_pixels(pattern,n=None,fraction=None,seed=None):
     # sorting the n/fraction kwargs
 
     both_none = n is None and fraction is None
-    neither_none =  n is not None and fraction is not None
+    neither_none = n is not None and fraction is not None
 
     if both_none or neither_none:
         raise ValueError("Exactly one of 'n' and 'fraction' must be None")
@@ -154,15 +158,15 @@ def add_dead_pixels(pattern,n=None,fraction=None,seed=None):
     rng = _process_seed_argument(seed)
 
     # .astype rounds down, these generate values from 0 to (pattern.shape - 1)
-    xdead = rng.uniform(low=0,high=pattern.shape[0],size=n).astype(int)
-    ydead = rng.uniform(low=0,high=pattern.shape[1],size=n).astype(int)
+    xdead = rng.uniform(low=0, high=pattern.shape[0], size=n).astype(int)
+    ydead = rng.uniform(low=0, high=pattern.shape[1], size=n).astype(int)
 
-    pattern[xdead,ydead] = 0
+    pattern[xdead, ydead] = 0
 
     return pattern
 
 
-def add_linear_detector_gain(pattern,gain):
+def add_linear_detector_gain(pattern, gain):
     """
     Multiplies the pattern by a gain (which is not a function of the pattern)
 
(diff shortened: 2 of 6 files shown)


dnjohnstone:
Thanks for this @pc494 , looks like a good structure - just to note that I think that this should incorporate and/or close #35

pc494:
Tidying up now, need to do:

- [x] pytest
- [x] black
- [x] coveralls

hakonanes:
~i already covered the seed = None case in the {hkl} PR... i can resolve conflicts after this is merged~

Sorry, wrong seed.

dnjohnstone:
@pc494 can we have a conclusion on incorporating #35 here before merging please - the code in that issue should be pretty much good to go, it just didn't have a natural home in the package before now.

pc494:
> @pc494 can we have a conclusion on incorporating #35 here before merging please - the code in that issue should be pretty much good to go, it just didn't have a natural home in the package before now.

So I've had a look at this now, I think adding `gain` and `offset` and `speckle` as functions makes a lot of sense, and would be sufficient to close #35 

But that got me thinking that I may as well complete the job (as it were) which I think involves also splitting the "gaussians" into 

`gaussian_spatial `("smearing")
`gaussian_intensity` ("detector error")

If I can find a source for why the intensity distribution might have a gamma error I'll add that too.

diffsims/pattern/detector_functions.py
```diff
@@ -16,13 +16,48 @@
...
+    """ Sets up a numpy random number generator with a seed"""
+    if seed is not None:
+        rng = default_rng(seed)
+    else:
+        rng = default_rng()
+
+    return rng
+
+
+def constrain_to_dynamic_range(pattern, detector_max=None):
```
hakonanes, line 35: BTW, this can be handled by `skimage.exposure.rescale_intensity`, I believe: https://scikit-image.org/docs/stable/api/skimage.exposure.html#skimage.exposure.rescale_intensity.

pc494: I did think skimage would have some of this code built in, but I've found that they can have a slightly strange set of habits about dtypes. Eg, that's why I've used numpy for my shot noise (which should be integer imo) code rather than the built skimage code. 

Also, as it happens the linked function does something slightly different to what we need. A value of 200 (detector_max set to 255) should be unchanged (for our purposes) whereas skimage would (if 200 was the max in the image) rescale it up to 255

hakonanes: I agree that skimage doesn't have the solution all the time.

And I think I misunderstood your intent: you don't want to rescale anything, just clip values. Then you don't want that function. Sorry for the... noise.

hakonanes:
Will review this tomorrow morning, also trying it out on some simulated EBSD patterns with no noise to see if we can use this to simulate e.g. detector gain!

dnjohnstone:
Ok, I agree with all of the points you made - I was more interested in making sure we had a mixed poisson/gaussian noise option and had forgotten about the gamma noise thing, which I agree isn't important. On this bit though....

> But that got me thinking that I may as well complete the job (as it were) which I think involves also splitting the "gaussians" into
> 
> `gaussian_spatial `("smearing")
> `gaussian_intensity` ("detector error")

perhaps we want to think a little about terminology here? I usually went back to this paper for that: https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI)1097-0029(20000501)49:3%3C245::AID-JEMT4%3E3.0.CO;2-O (although there are others and Marc DeGraef's "conventional transmission electron microscopy" is good on this topic too.

My point being - is "gaussian_spaital" = "point_spread_function" and "gaussian_intensity" = "gaussian_noise"?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - API reference documentation via Read The Docs: https://diffsims.rtfd.io
 - New module: `sphere_mesh_generators`
+- New module: `detector_fucntions`
```
pc494: typo. oops.

pc494:
> I was more interested in making sure we had a mixed poisson/gaussian noise option 

In the short term you can do this by stacking them one after the other (I don't think the operations commute so I'm not inclined to bundle them together myself)

> My point being - is "gaussian_spaital" = "point_spread_function" and "gaussian_intensity" = "gaussian_noise"?

So I ended up having the names as `add_gaussian_blur` and `add_gaussian_noise`. I think `add_gaussian_noise` is pretty clear and fits this module nicely.

I agree the blurring is a bit different, it's certainly like a point spread function, although I had been thinking of it as an MTF but that might be my mistake. In future we could think about separating the spatial and intensity detector functions? Thoughts are welcome.

dnjohnstone:
> In the short term you can do this by stacking them one after the other (I don't think the operations commute so I'm not inclined to bundle them together myself)

They don't commute, but isn't that a reason to bundle them to avoid people doing them in the wrong order? I think it's almost always Poisson first for electron detection?

> I agree the blurring is a bit different, it's certainly like a point spread function, although I had been thinking of it as an MTF but that might be my mistake. In future we could think about separating the spatial and intensity detector functions?

I think if you're doing a convolution with a Gaussian then the Gaussian is a point spread function- if you're just multiplying the image by a Gaussian then it would be and MTF. I didn't check very hard, but I assume you're doing the former? https://en.wikipedia.org/wiki/Point_spread_function

pc494:
> > In the short term you can do this by stacking them one after the other (I don't think the operations commute so I'm not inclined to bundle them together myself)
> 
> They don't commute, but isn't that a reason to bundle them to avoid people doing them in the wrong order? I think it's almost always Poisson first for electron detection?

In terms of detector errors, yes, so I will write that function.

> I think if you're doing a convolution with a Gaussian then the Gaussian is a point spread function- if you're just multiplying the image by a Gaussian then it would be and MTF. I didn't check very hard, but I assume you're doing the former? https://en.wikipedia.org/wiki/Point_spread_function

I am doing the former (ie. PSF), and I think I need to reread what an MTF does.

pc494:
So an MTF (really a special case of the Optical Transfer Function) is the fourier transform of a point spread function. And `gaussian_blur` is indeed giving us the detector point spread function (I'll link the wiki page in the reference section).

Hoping to finish this PR up today.

pc494:
Okay, @hakonanes (or others) this is now good for review.

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+    pattern[pattern < 0] = 0
+
+    if detector_max is not None:
+        pattern[pattern > detector_max] = detector_max
+
+    return pattern
+
+
+def add_gaussian_blur(pattern, sigma):
```
dnjohnstone: for consistency with other parts of this PR now - should this be `add_gaussian_point_spread`?

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,240 @@
...
+    pattern intensity, so be mindful to put your intensities in physical units
+    """
+    rng = _process_seed_argument(seed)
+
+    return rng.poisson(pattern)
+
+
+def add_shot_and_point_spread(pattern, sigma, shot_noise=True, seed=None):
+    """
+    Adds (potentially) both shot and point spread noise to a pattern
```
dnjohnstone: To be more explicit and clarify that this is a gaussian point spread - with implications on symmetry in x/y etc, how about:

Adds shot noise and/or convolution with a gaussian point spread function to a pattern.

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,240 @@
...
+        seed value for the random number generator (effects the shot noise only)
+
+    Returns
+    -------
+    detector_pattern : np.array
+        A single sample of the pattern after accounting for detector properties
+
+    See also
+    --------
+    add_shot_noise : adds only shot noise
```
dnjohnstone, line 125: also add_gaussian_point_spread

dnjohnstone (changes requested):
Basically looks good - a couple of minor naming things suggested.

dnjohnstone (approved):

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,241 @@
...
+        pattern_size = pattern.shape[0] * pattern.shape[1]
+        n = int(fraction * pattern_size)
+
+    rng = _process_seed_argument(seed)
+
+    # .astype rounds down, these generate values from 0 to (pattern.shape - 1)
+    xdead = rng.uniform(low=0, high=pattern.shape[0], size=n).astype(int)
+    ydead = rng.uniform(low=0, high=pattern.shape[1], size=n).astype(int)
+
+    pattern[xdead, ydead] = 0
```
hakonanes, line 201: Should perhaps call these (rowdead, coldead)? As it stands now, xdead are the rows.

hakonanes, line 201: This shouldn't operate inplace, but on new copy of the pattern (deepcopy()).

I think this should be the case for all functions.

hakonanes (changes requested):
Looks very good overall! Looking forward to use these functions to simulate gain etc. and see how our denoising methods in kikuchipy can "retain" the signal.

The functions should return a new copy of the pattern, not operate inplace, right? I was surprised the pattern operated inplace, that is the input pattern after passed is identical to the returned pattern...

hakonanes:
Sorry for being late, but I hope that the functions can return a new copy instead of operating inplace.

pc494:
Good review, I will deal with this in a PR tomorrow and make a new release candidate.

diffsims/pattern/detector_functions.py
pc494: I think this bug only happens in this function, but I will write tests for all of them and fix this

pc494: I think x is normally our row direction?

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,241 @@
...
+    """ Sets up a numpy random number generator with a seed"""
+    if seed is not None:
+        rng = default_rng(seed)
+    else:
+        rng = default_rng()
+
+    return rng
+
+
+def constrain_to_dynamic_range(pattern, detector_max=None):
```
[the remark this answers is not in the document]
pc494: I had kind of expect some users to make use of it, so was planning on it being public. For example you might have 1e5 signal, that you want to apply a point spread too and then constrain to the dynamic range.

diffsims/pattern/detector_functions.py
```diff
@@ -0,0 +1,241 @@
...
+    """Force the values within pattern to lie between [0,detector_max]
+
+    Parameters
+    ----------
+    pattern : np.array
```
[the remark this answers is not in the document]
pc494: sweet, will implement this adjustment

diffsims/pattern/detector_functions.py
hakonanes: you mean x increases vertically? i usually think of x from left to right and y from up to down. i guess it has no consequence for the user, so i dont mind this choice

hakonanes: np.may_share_memory(array1, array2) returning true/false is a test i use often.

pc494: I was getting confused by the word "row", this is the wrong way round and is being fixed in the upcoming PR

PR #145, opened 2020-12-15
Make ReciprocalLatticePoint handle having only one point/vector

### LINKED ISSUE #125: ReciprocalLatticePoint handles the case when it contains only one point badly
**Describe the bug**
Getting allowed reflections from ReciprocalLatticePoint.allowed with only one point fails.

**To Reproduce**
```python
>>> from diffsims.crystallography import ReciprocalLatticePoint
>>> from orix.crystal_map import Phase
>>> rlp = ReciprocalLatticePoint(phase=Phase(), hkl=[1, 1, 1])
>>> rlp.allowed
Traceback (most recent call last):
[...]  # Unrelevant output
  File "<ipython-input-32-0c7dae704f84>", line 1, in <module>
    rlp2.allowed
  File "/home/hakon/miniconda3/envs/kp-dev/lib/python3.8/site-packages/diffsims/crystallography/reciprocal_lattice_point.py", line 159, in allowed
    selection = np.sum(np.mod(self._hkldata, 2), axis=1)
  File "<__array_function__ internals>", line 5, in sum
  File "/home/hakon/miniconda3/envs/kp-dev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2241, in sum
    return _wrapreduction(a, np.add, 'sum', axis, dtype, out, keepdims=keepdims,
  File "/home/hakon/miniconda3/envs/kp-dev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 87, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
numpy.AxisError: axis 1 is out of bounds for array of dimension 1
```

The problem is most likely the `rlp._hkldata` property (which is a convenience property returning hkl's as a numpy array) which squeezes out 1-dimensions, thus removing the effect of `np.atleast_2d(hkl)` which `Vector3d` imposes upon initialization...

Will fix this and add a test for single point behaviour.

hakonanes:
Signed-off-by: Håkon Wiik Ånes <hwaanes@gmail.com>

---
name: Make ReciprocalLatticePoint handle having only one point/vector
about: Fix #125 

---

**Release Notes**
> bugfix
ReciprocalLatticePoint handles having only one point/vector

* Close #125.
* Make `ReciprocalLatticePoint` handle having only one point/vector by not using the `_hkldata` attribute. The introduction of that attribute was not well thought through, so I removed it. A test is added to ensure that `ReciprocalLatticePoint.allowed` works with only one vector.
* Add private methods to raise a `ValueError` if the `ReciprocalLatticePoint` requires a point or space group, but it's set to `None`. And use these in `init()` and the `allowed` property, respectively. Anticipating that this will be used more, therefore private methods.

kikuchipy uses the `ReciprocalLatticePoint._hkldata`, so we have to change this use to `.hkl.data`!

#### For reviewers

  later.

### DIFF
--- a/diffsims/crystallography/reciprocal_lattice_point.py
+++ b/diffsims/crystallography/reciprocal_lattice_point.py
@@ -18,6 +18,7 @@
 
 from collections import defaultdict
 from itertools import product
+import warnings
 
 import numpy as np
 from orix.vector import Vector3d
@@ -51,8 +52,9 @@ class ReciprocalLatticePoint:
         hkl : orix.vector.Vector3d, np.ndarray, list, or tuple
             Miller indices.
         """
-        self._hkl = Vector3d(hkl)
         self.phase = phase
+        self._raise_if_no_point_group()
+        self._hkl = Vector3d(hkl)
         self._structure_factor = [None] * self.size
         self._theta = [None] * self.size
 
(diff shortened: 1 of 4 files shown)


pc494:
Will review this this afternoon :)

diffsims/crystallography/reciprocal_lattice_point.py
```diff
@@ -360,6 +356,20 @@ def calculate_theta(self, voltage):
         wavelength = get_refraction_corrected_wavelength(self.phase, voltage)
         self._theta = np.arcsin(0.5 * wavelength * self.gspacing)
 
+    def _raise_if_no_point_group(self):
```
pc494, line 359: How often do you anticipate using these methods? I would be tempted to either write the code explicitly where it's used, or rename them `check_point_group_present` or similar? Might need more thinking time on it.

pc494 (commented):
On the fence about the commented bit, will see how I feel in a day or two.

pc494 (approved):
Had a change of heart, I was being too picky, will merge.

PR #146, opened 2020-12-15
Get {hkl} families from a list of (hkl) in the ReciprocalLatticePoint class

### LINKED ISSUE #133: Convenience method to find HKLs in a ReciprocalLatticePoint symmetrically equivalent to HKL
This is more of a note to myself than an issue or problem, but feel free to comment.

When wanting to color all Kikuchi bands belonging to the {111} family red (https://github.com/pyxem/kikuchipy/pull/232/files\#diff-c924e24a2150b7bd8c6b13646815e33791443b78c109e752306e2cf31c99ce15R488, see it in action here: https://github.com/hyperspy/hyperspy/pull/2558#issue-499539271), I found myself writing code which should be a method of the `diffsims.crystallography.ReciprocalLatticePoint` class.

The idea is to:
1. Get unique families in `ReciprocalLatticePoint`
2. Get the bands corresponding to each family (and e.g. color them)

I think the simplest syntax would be (will get proper syntax later)

```python
>>> rlp.families  # Class property, list symmetrically unique {hkl}
>>> rlp.families[0]
[1 1 1]  # E.g.
>>> rlp.families_contain  # Class property, list families and hkl in that family
>>> rlp.is_equivalent(rlp.families[0])  # Returns a boolean array
```

I would like this in as part of a v0.4, along with a fix for #125 and some other things into that class that I'll get to.

hakonanes:
Signed-off-by: Håkon Wiik Ånes <hwaanes@gmail.com>

---
name: Get families {hkl} in list of (hkl)
about: Add families attribute to ReciprocalLatticePoint

---

**Release Notes**
> new feature
Add possibility to get families {hkl} from list of (hkl) `ReciprocalLatticePoint`

Close #133 

Add the following `ReciprocalLatticePoint` class attributes:
* `.families`: array of {hkl} families from all (hkl) in the instance
* `.n_families`: number of {hkl} families
* `.family`: index into `families` array for each (hkl)

These new attributes are calculated from the new or updated functions:
* `get_hkl_family()` (new): Return {hkl} families and to which family each (hkl) belongs. Public function, importable from `diffsims.crystallography.get_hkl_family`.
* `_reduce_hkl()` (new): Reduce Miller indices 3-tuples by a greatest common divisor.
* `_is_equivalent()` (updated): Determine whether two sets of Miller indices are equivalent. Symmetry is not considered.

Tests are added for this functionality, but only for the cubic crystal system...

With the addition of `_reduce_hkl()`, a `reduce: bool = False` parameter is added to `ReciprocalLatticePoint.unique()` so that e.g. [[2, 0, 0], [1, 0, 0]] -> [1, 0, 0].

Other additions:
* TODOs to remove `get_unique_families()` and corresponding test.
* Minor docstring intersphinx corrections
* Added test of `seed` parameter in `get_random_sphere_vertices()` -> 100% coverage.

**Example**

```python
>>> from orix.crystal_map import Phase
>>> from diffsims.crystallography import ReciprocalLatticePoint
>>> rlp = ReciprocalLatticePoint(
...     phase=Phase(space_group=225),
...     hkl=[[-1, -1, -1], [1, 0, 0], [1, 1, 1], [0, 0, 1], [2, 0, 0]
... )
>>> rlp
ReciprocalLatticePoint (5,)
Phase: ferrite (m-3m)
[[-1 -1 -1]
 [ 1  0  0]
 [ 1  1  1]
 [ 0  0  1]
 [ 2  0  0]]
>>> rlp.unique()
ReciprocalLatticePoint (3,)
Phase: ferrite (m-3m)
[[-1 -1 -1]
 [ 1  0  0]
 [ 2  0  0]]
>>> rlp.unique(reduce=True)
ReciprocalLatticePoint (2,)
Phase: ferrite (m-3m)
[[-1 -1 -1]
 [ 1  0  0]]
>>> rlp.n_families
2
>>> rlp.families
array([[-1, -1, -1],
       [ 1,  0,  0]])
>>> rlp.family
array([0, 1, 0, 1, 1])
```

This functionality is in the [kikuchipy.crystallography module](https://github.com/pyxem/kikuchipy/blob/master/kikuchipy/crystallography/_computations.py#L46), but belongs here.

**Work in progress?**
- [x] Merge `master` with #145 into this branch
- [x] Add actual `ReciprocalLatticePoint.families` attribute
- [x] Test it (port [tests from kikuchipy](https://github.com/pyxem/kikuchipy/blob/master/kikuchipy/crystallography/tests/test_crystallographic_computations.py#L144))
- [x] Add changelog entry

#### For reviewers

  later.

### DIFF
--- a/diffsims/crystallography/reciprocal_lattice_point.py
+++ b/diffsims/crystallography/reciprocal_lattice_point.py
@@ -89,6 +89,7 @@ class ReciprocalLatticePoint:
         """Return :class:`numpy.ndarray` without 1-dimensions."""
         return np.squeeze(self.hkl.data)
 
+    @property
     def h(self):
         """Return :class:`np.ndarray` of Miller index h."""
         return self.hkl.data[..., 0]
--- a/diffsims/crystallography/reciprocal_lattice_point.py
+++ b/diffsims/crystallography/reciprocal_lattice_point.py
@@ -48,7 +48,8 @@ class ReciprocalLatticePoint:
         phase : orix.crystal_map.phase_list.Phase
             A phase container with a crystal structure and a space and
             point group describing the allowed symmetry operations.
-        hkl : orix.vector.Vector3d, np.ndarray, list, or tuple
+        hkl : orix.vector.vector3d.Vector3d, numpy.ndarray, list, or\
+                tuple
             Miller indices.
         """
         self._hkl = Vector3d(hkl)
@@ -77,27 +78,29 @@ class ReciprocalLatticePoint:
 
     @property
     def hkl(self):
-        """Return :class:`~orix.vector.Vector3d` of Miller indices."""
+        """Return :class:`~orix.vector.vector3.Vector3d` of Miller
+        indices.
+        """
         return Vector3d(self._hkl.data.astype(int))
 
     @property
     def _hkldata(self):
-        """Return :class:`np.ndarray` without 1-dimensions."""
+        """Return :class:`numpy.ndarray` without 1-dimensions."""
         return np.squeeze(self.hkl.data)
 
     @property
     def h(self):
-        """Return :class:`np.ndarray` of Miller index h."""
+        """Return :class:`numpy.ndarray` of Miller index h."""
         return self._hkldata[..., 0]
 
     @property
     def k(self):
-        """Return :class:`np.ndarray` of Miller index k."""
+        """Return :class:`numpy.ndarray` of Miller index k."""
         return self._hkldata[..., 1]
 
     @property
     def l(self):
-        """Return :class:`np.ndarray` of Miller index l."""
+        """Return :class:`numpy.ndarray` of Miller index l."""
         return self._hkldata[..., 2]
 
     @property
@@ -112,31 +115,31 @@ class ReciprocalLatticePoint:
 
     @property
     def multiplicity(self):
-        """Return either `int` or :class:`np.ndarray` of `int`."""
+        """Return either `int` or :class:`numpy.ndarray` of `int`."""
         return self.symmetrise(antipodal=True, return_multiplicity=True)[1]
 
     @property
     def gspacing(self):
-        """Return :class:`np.ndarray` of reciprocal lattice point
+        """Return :class:`numpy.ndarray` of reciprocal lattice point
         spacings.
         """
         return self.phase.structure.lattice.rnorm(self._hkldata)
 
     @property
     def dspacing(self):
-        """Return :class:`np.ndarray` of direct lattice interplanar
+        """Return :class:`numpy.ndarray` of direct lattice interplanar
         spacings.
         """
         return 1 / self.gspacing
 
     @property
     def scattering_parameter(self):
-        """Return :class:`np.ndarray` of scattering parameters s."""
+        """Return :class:`numpy.ndarray` of scattering parameters s."""
         return 0.5 * self.gspacing
 
     @property
     def structure_factor(self):
-        """Return :class:`np.ndarray` of structure factors F or None."""
+        """Return :class:`numpy.ndarray` of structure factors F or None."""
         return self._structure_factor
 
     @property
@@ -171,8 +174,12 @@ class ReciprocalLatticePoint:
 
     @property
     def theta(self):
-        """Return :class:`np.ndarray` of twice the Bragg angle."""
+        """Return :class:`numpy.ndarray` of twice the Bragg angle."""
         return self._theta
+
+    @property
+    def families(self):
+        """Return symmetrically unique Miller indices."""
 
     @classmethod
     def from_min_dspacing(cls, phase, min_dspacing=0.5):
@@ -204,7 +211,7 @@ class ReciprocalLatticePoint:
         phase : orix.crystal_map.phase_list.Phase
             A phase container with a crystal structure and a space and
             point group describing the allowed symmetry operations.
-        highest_hkl : np.ndarray, list, or tuple of int
+        highest_hkl : numpy.ndarray, list, or tuple of int
             Highest Miller indices to consider (including).
         """
         hkl = get_hkl(highest_hkl=highest_hkl)
@@ -272,7 +279,7 @@ class ReciprocalLatticePoint:
         ReciprocalLatticePoint
             Planes with Miller indices symmetrically equivalent to the
             original planes.
-        multiplicity : np.ndarray
+        multiplicity : numpy.ndarray
             Multiplicity of the original Miller indices. Only returned if
             `return_multiplicity` is True.
 
@@ -375,7 +382,7 @@ def get_highest_hkl(lattice, min_dspacing=0.5):
 
     Returns
     -------
-    highest_hkl : np.ndarray
+    highest_hkl : numpy.ndarray
         Highest Miller indices.
     """
     highest_hkl = np.ones(3, dtype=int)
@@ -394,12 +401,13 @@ def get_hkl(highest_hkl):
 
     Parameters
     ----------
-    highest_hkl : orix.vector.Vector3d, np.ndarray, list, or tuple of int
+    highest_hkl : orix.vector.vector3d.Vector3d, numpy.ndarray, list,\
+            or tuple of int
         Highest Miller indices to consider.
 
     Returns
     -------
-    hkl : np.ndarray
+    hkl : numpy.ndarray
         An array of Miller indices.
     """
     index_ranges = [np.arange(-i, i + 1) for i in highest_hkl]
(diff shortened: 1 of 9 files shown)


dnjohnstone:
Thanks @hakonanes - this will be a good addition. It probably should deprecate the `get_unique_families` util function https://github.com/pyxem/diffsims/blob/9a480149097ca6794ed4f447f5cd84faa446e285/diffsims/utils/sim_utils.py#L76 and that will have a knock on impact on the 1D diffraction pattern simulator following #102 

That was all a fairly bodged part of the code-base, so really good to improve through the to be `ReciprocalLatticeVector` class.

hakonanes:
OK, thanks a lot for letting me know. I will have a look at that functionality and see how to replace it without causing any disruption. Perhaps in another PR (patch?), will see today and tomorrow.

pc494:
I'm happy to tidy up in the release sweep, as I've seen that code before.

diffsims/crystallography/reciprocal_lattice_point.py
```diff
@@ -464,5 +495,73 @@ def get_equivalent_hkl(hkl, operations, unique=False, return_multiplicity=False)
...
+    hkl = hkl[~np.all(np.isclose(hkl, 0), axis=1)]
+    families = defaultdict(list)
+    families_idx = defaultdict(list)
+    for i, this_hkl in enumerate(hkl.tolist()):
+        for that_hkl in families.keys():
+            if _is_equivalent(this_hkl, that_hkl, reduce=reduce):
+                families[tuple(that_hkl)].append(this_hkl)
+                families_idx[tuple(that_hkl)].append(i)
+                break
+        else:
```
pc494, line 522: this indentation looks off to me?

diffsims/crystallography/reciprocal_lattice_point.py
```diff
@@ -464,5 +495,73 @@ def get_equivalent_hkl(hkl, operations, unique=False, return_multiplicity=False)
...
+            families[tuple(this_hkl)].append(this_hkl)
+            families_idx[tuple(this_hkl)].append(i)
+    return families, families_idx
+
+
+def _is_equivalent(this_hkl, that_hkl, reduce=False):
```
pc494, line 528: Could you explain this function to me?

diffsims/tests/test_generators/test_sphere_mesh_generators.py
```diff
@@ -36,6 +37,16 @@ def test_random_sphere_mesh():
     assert grid.shape[0] == 10313
     assert grid.shape[1] == 3
 
+    res = 10
```
pc494, line 40: this is the bit that will have a conflict with #143 - should be simple to resolve

pc494 (commented):
Couple of questions, but broadly looks good.

dnjohnstone:
Am I missing something or will this only work for cubic systems?

hakonanes:
> Am I missing something or will this only work for cubic systems?

I've only tested this ~class~ functionality for cubic crystal structures, so that may be the case. Could you give me an example of two symmetrically equivalent planes in another crystal system so that I can add that to the tests?

dnjohnstone:
> > Am I missing something or will this only work for cubic systems?
> 
> I've only tested this ~class~ functionality for cubic crystal structures, so that may be the case. Could you give me an example of two symmetrically equivalent planes in another crystal system so that I can add that to the tests?

Well the easiest one to test is the negative i.e. (100) not equal to (010) not equal to (001) - for anything other than cubic (and ~orthorhombic~ tetragonal, where the first one will be equal), for example.

hakonanes:
> Well the easiest one to test is the negative i.e. (100) not equal to (010) not equal to (001) - for anything other than cubic (and orthorhombic, where the first one will be equal), for example.

~That case is taken care of in the tests: https://github.com/pyxem/diffsims/pull/146/files#diff-54c94aa2aea561781caf79975ba8bd988195449a3d581e0e88fbdaff205f7b62R293~

~Edit: permutations are taken care of, as well as higher order.~

Sorry, that is the opposite of what you want. So yes, the `ReciprocalLatticePoint.unique()` method and these new properties only give the correct result for cubic crystals.

hakonanes:
Do I understand it correctly if current implementations https://github.com/pyxem/diffsims/blob/9a480149097ca6794ed4f447f5cd84faa446e285/diffsims/utils/sim_utils.py#L76

also only gives a desired result for cubic crystals?

pc494:
> Do I understand it correctly if current implementations
> 
> https://github.com/pyxem/diffsims/blob/9a480149097ca6794ed4f447f5cd84faa446e285/diffsims/utils/sim_utils.py#L76
> 
> also only gives a desired result for cubic crystals?

Yes, I think so, my confusion is due to the docstrings of `_is_equivalent` says something to the effect of "doesn't consider symmetry" but in fact it's working under cubic symmettry. In the triclinic system (no sym) you would have

(0,0,1) != (0,1,0) != (1,0,0) etc.

Also just to clear up Duncan's remarks:

> Well the easiest one to test is the negative i.e. (100) not equal to (010) not equal to (001) - for anything other than cubic (and orthorhombic, where the first one will be equal), for example.

This is a typo I think, "orthorhombic" should read "tetragonal":

Cubic: (100) equivalent to (010) and to (001)
Tetragonal: (100) equivalent to (010) but different from (001) _using standard settings_

diffsims/tests/test_crystallography/test_reciprocal_lattice_point.py
```diff
@@ -286,6 +286,26 @@ def test_calculate_theta(self, ferrite_phase, voltage, hkl, desired_theta):
         rlp.calculate_theta(voltage=voltage)
         assert np.allclose(rlp.theta, desired_theta)
 
+    @pytest.mark.parametrize(
+        "hkl, family_heads, family_idx, n_families",
+        [
+            (
+                [[1, 1, 1], [1, 0, 0], [-1, -1, -1], [0, 1, 0], [2, 0, 0]],
+                [[1, 1, 1], [1, 0, 0]],
```
dnjohnstone, line 294: @hakonanes  my point is that if the system is hexagonal, orthorhombic, monoclinic or triclinic - the correct answer here should be family_heads=[[1,1,1], [1,0,0], [0,1,0]] - I don't think that's currently tested for? Or what the code would return?

In response to your other question, it's quite possible that the current code only works for cubic systems... but always looking to improve, we should at least log it as an issue if that is the case...

dnjohnstone:
> This is a typo I think, "orthorhombic" should read "tetragonal":
> 
> Cubic: (100) equivalent to (010) and to (001)
> Tetragonal: (100) equivalent to (010) but different from (001) _using standard settings_

whoops, yes, that's what I meant, thanks!

diffsims/tests/test_crystallography/test_reciprocal_lattice_point.py
dnjohnstone: Oh, and I've just seen your update to your comment - I think we're all on the same page now :)

hakonanes: The point group symmetry of the `ReciprocalLatticePoint` instance is only considered in the `.symmetrise()` method, not in the `.unique()` method, thus the logic in the latter method should only works for cubic crystals as it is implemented... We should fix this, see my comment below.

hakonanes:
I've decided I don't want to merge this until I've added logic and tests for all crystal systems... @pc494, I guess this means this will have to wait until a 0.4.1, since I consider this a bug?

pc494:
I'm happy to make a 0.4.0 that doesn't have this in it. We can then decide if we want the addition to be considered as 0.4.1 or 0.5.0, either seems reasonable to me.

hakonanes:
yeah, this class needs a more general approach, because i want to use it to simulate geometrical and kinematical ebsd patterns. so i'm happy to wait to improve this, which can either be a 0.5 or  0.4.1, i think.

hakonanes:
I plan to replace most of the `ReciprocalLatticePoint` class with a `orix.vector.Miller` class, while also fixing the symmetrically equivalent plane issues.

pc494:
I've been thinking about making a 0.4.1 patch for diffsims in the next couple of days. Is it fair to say this is more of a minor release PR?

hakonanes:
Great, @pc494! Yes, I'm currently working on a general implementation in https://github.com/pyxem/orix/pull/164, and will then continue on this PR to inherit from `orix.vector.Miller`.

pc494:
Cool, that's broadly what I thought was going on, just wanted to make sure.

hakonanes:
This will be superseded by a new PR which implements a new `ReciprocalLatticeVector` class to replace `ReciprocalLatticePoint`.

PR #154, opened 2021-03-03
added pattern-to-mask feature

din14970:
---
name: Template to binary mask
about: Implemented a small feature whereby a template can be converted to corresponding binary mask of circles.

---

**Release Notes**
* Added a method `get_as_mask` in `DiffractionSimulation` with a number of options
* Added a module `mask_utils` under utils. 

For my work I would like to easily generate masks from templates to perform multi-stage indexations. I always implemented this in hacky ways but decided it would be handy if `DiffractionSimulation` would have a method that converts it to a binary mask, similar as it can be converted to an image of Gaussians.

A simple example:
```python
import diffpy

aus = diffpy.structure.loadStructure("austeniteLP.cif")
diff_gen = DiffractionGenerator(accelerating_voltage=200, shape_factor_model="lorentzian")

orientation = [0, 20, 45]
simulation = diff_gen.calculate_ed_data(aus,
                                        2.3,
                                        orientation,
                                        with_direct_beam=False,
                                        max_excitation_error=1e-1,
                                        )
simulation.calibration = 0.012

mask = simulation.get_as_mask((256, 256), radius = 10)
```
Plotting the simulated coordinates produces:
afbeelding

The image of the mask: 
afbeelding

One can specify whether one wants a "negative" or a "positive" mask. The radii of the circles can also be controlled - each can be set individually by passing an array into `radius` or the radius can be determined based on a function of the intensity, which can be passed in as `radius_function`.

Modifications can be made to the mask using additional functions in `mask_utils`. I added support for drawing a circle, any polygon, and an annulus. The functions work in-place, e.g.

```python
from pyxem.utils import mask_utils as mu
mask = mu.create_mask((512, 512))
mu.add_annulus_to_mask(150, 250, mask)
mu.add_circle_to_mask(256, 50, 30, mask, fill=True)
coords = np.array([[200, 200],
                   [350, 100],
                   [250, 350]])
mu.add_polygon_to_mask(coords, mask)
```
This produces a mask:
afbeelding

Maybe I didn't search hard enough, but I couldn't find these convenience functions to generate a mask in other packages.

I'm unsure whether the `mask_utils` fit all that well in diffsims and they would probably fit better in pyxem. On the other hand, I like that `get_as_mask` function being under `DiffractionSimulation`.

I would say it's fully functional but a discussion may be needed on where to best place things. Also unit tests are still needed.

A lazy implementation using e.g. dask.futures or dask.delayed might be handy to generate a 4D-STEM sized mask without having it in memory. 

**Work in progress?**
- [x] Unit testing
- [ ] Find a good place for mask functions

#### For reviewers

  later.

### DIFF
--- a/diffsims/sims/diffraction_simulation.py
+++ b/diffsims/sims/diffraction_simulation.py
@@ -19,6 +19,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from diffsims.pattern.detector_functions import add_shot_and_point_spread
+from diffsims.utils import mask_utils
 
 
 class DiffractionSimulation:
@@ -123,6 +124,45 @@ class DiffractionSimulation:
     @intensities.setter
     def intensities(self, intensities):
         self._intensities = intensities
+
+    def get_as_mask(self, shape, radius=6., negative=True,
+                    radius_function=None, direct_beam_position=None,
+                    *args, **kwargs):
+        """
+        Return the diffraction pattern as a binary mask of type
+        bool
+
+        Parameters
+        ----------
+        shape: 2-tuple of ints
+            Shape of the output mask (width, height)
+        radius: float or array, optional
+            Radii of the spots in pixels. An array may be supplied
+            of the same length as the number of spots.
+        negative: bool, optional
+            Whether the spots are masked (True) or everything
+            else is masked (False)
+        radius_function: Callable, optional
+            Calculate the radius as a function of the spot intensity,
+            for example np.sqrt. args and kwargs supplied to this method
+            are passed to this function. Will override radius.
+        direct_beam_position: 2-tuple of ints, optional
+            The (x,y) coordinate in pixels of the direct beam. Defaults to
+            the center of the image.
+        """
+        r = radius
+        cx, cy = shape[0]//2, shape[1]//2
+        if direct_beam_position is not None:
+            cx, cy = direct_beam_position
+        point_coordinates_shifted = self.calibrated_coordinates[:, :-1].copy()
+        point_coordinates_shifted[:, 0] += cx
+        point_coordinates_shifted[:, 1] += cy
+        if radius_function is not None:
+            r = radius_function(self.intensities, *args, **kwargs)
+        mask = mask_utils.create_mask(shape, fill=negative)
+        mask_utils.add_circles_to_mask(point_coordinates_shifted, r, mask,
+                                    fill=not negative)
+        return mask
 
     def get_diffraction_pattern(self, size=512, sigma=10):
         """Returns the diffraction data as a numpy array with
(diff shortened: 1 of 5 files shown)


hakonanes:
Without looking at the code, I quite like the masking utilities you're showing, and the functionality your describing. I would like something similar for extracting Kikuchi band intensities from positions of geometrical EBSD simulations in kikuchipy, and it looks like this is a perfect platform to build that from. I strongly suggest this be placed in diffsims and not in pyxem.

I'll leave it for someone else to review it, as I haven't used that part of the code.

din14970:
@hakonanes I suppose this could be added as a utility function like `mask_band(xy, in-plane_angle, width)`. Shouldn't be too complicated.

din14970:
A band feature on the mask is also possible now:

image

I'll just create some tests for these things

din14970:
@pc494 I would say this is ready for review

pc494 (changes requested):
This basically looks good to me apart from one thing.

It seems you are fixing masks inplace (which is fine), could you make this explicit in the various bits of docstrings:

eg: 
```
Returns
----------
mask : etc.
```

or 
```
Returns
----------
None : 
     mask is adjusted inplace
```

hakonanes:
> It seems you are fixing masks inplace (which is fine), could you make this explicit in the various bits of docstrings:

If operations are done inplace I *always* state this in the top docstring description, e.g. ["Remove the static background in an EBSD scan inplace"](https://kikuchipy.org/en/latest/reference.html#kikuchipy.signals.EBSD.remove_static_background).

pc494:
@din14970 if you are happy to deal with:
`Check that new functions are imported in corresponding __init__.py.`

I'll sort out the CHANGELOG, as it's got some conflicts at the moment after the patch release came out

din14970:
@pc494  Hope that this is acceptable. Sorry for my ignorance but could you explain what the purpose of adding it to the `__init__` file is? Just so that all the functions are directly available under the `utils` namespace or?

hakonanes:
> @pc494 Hope that this is acceptable. Sorry for my ignorance but could you explain what the purpose of adding it to the `__init__` file is? Just so that all the functions are directly available under the `utils` namespace or?

I think a clean import structure is *very* important. Ideally, everything should be available via a `diffsims.module.Class` or `diffsims.module.function`. To get this, we need to make stuff easily available via `module/__init__.py`.

For orix e.g, we can do `from orix.vector import Vector3d`, instead of `from orix.vector.vector3d import Vector3d`, since `from orix.vector.vector3d import Vector3d` is placed in `vector/__init__.py`.

A final comment on this: we should add a todo-point in the PR template to ensure that we also add new functionality in the API reference documentation (doc/reference.rst). Will have a look at this in a later PR.

hakonanes:
Just to clarify: I think a clean and understandable import structure documented in the [API reference](https://diffsims.readthedocs.io/en/latest/reference.html) makes it easy for the user to quickly navigate the package, without having to rely on a user guide/notebook.

din14970:
@hakonanes @pc494 ok, thanks for the clarification! One thing to keep in mind then is that we don't duplicate names in different modules. Will take care about changelog and init next time. Hope these mask tools end up being useful, I often find creating a custom mask a quite annoying task.

hakonanes:
> One thing to keep in mind then is that we don't duplicate names in different modules

Good point, a clear and easily available import structure should reduce the chances of doing this!

PR #165, opened 2021-04-29
Explicity state Python 3.9 support, increase minimal version of NumPy to >= 1.17

### LINKED ISSUE #164: Support Python 3.9
I suggest these changes to the the build matrix:
* Use Python 3.8 and 3.9 (instead of 3.7 and 3.8)
* Add extra step for minimal versions, i.e. Python 3.6 and the minimal package versions stated in setup.py. Whenever this step fails, we can either try to support the older version by changing code, or increment the minimal version.

Including a build step for minimal versions led me to increase minimal supported package versions in [kikuchipy quite substantially](https://github.com/pyxem/kikuchipy/pull/348/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L117).

We should also update setup.py to reflect support for Python 3.8 and 3.9.

I can do all these things when I find the time.

hakonanes:
* Update setup.py to reflect support for Python 3.8 and 3.9
* Increase minimal version of NumPy to >= 1.17 because of use of `numpy.random.default_rng`
* I think minimal version of tqdm 0.4.9 was a typo, changed to 4.9
* Build only on Python 3.8 and 3.9, but add a minimal dependencies step on Python 3.6. Note that to get this step to work, compatible versions of minimal dependencies must be installed, or else `pip` raises an incompatibility error. These versions are then: `pip install diffpy.structure==3.0.0 matplotlib==3.3 numpy==1.17 orix==0.5.0 scipy==1.0 tqdm==4.9`
* Touch up README
* Update changelog

Close #164.

#### For reviewers

  later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.7, 3.8,3.9]
+        python-version: [3.8, 3.9]
+        include:
+          # Oldest supported version of main dependencies on Python 3.6
+          - os: ubuntu-latest
+            python-version: 3.6
+            OLDEST_SUPPORTED_VERSION: true
+            # orix requires matplotlib 3.3, matplotlib 3.3 requires numpy 1.15
+            # numba requires scipy==1.0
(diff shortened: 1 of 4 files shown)


pc494 (approved):
LGTM

Issue #163, opened 2021-04-24
Simulating templates like commercial program - findings and possible bugs


din14970:
I've been trying to replicate the behavior of ASTAR for template matching for a bit now. While everything works as intended I can get quite different results in the orientation map. I believe the origin is to be found in the different ways the templates are actually generated; different calculated intensities means different "best match" patterns.

So I did a small (non-exhaustive) reverse engineering exercise, simulating the [001] zone axis pattern of FCC Fe for different conditions in ASTAR and checking the resulting spot intensities. Below is a plot of the spot intensity versus the distance of the spot to the Ewald sphere (I hope I have the units correct because there is no indication of units in the file - but changing them by orders of magnitude did not appreciably influence the conclusion) for different values of excitation error (see legend, again no units). 

![image](https://user-images.githubusercontent.com/25320842/115952697-ca197a00-a4e7-11eb-838d-982ef96adb3b.png)

The point is that the intensity of the spots varies nearly perfectly linearly with the distance from the ewald sphere, and that the excitation error only seems to influence the width of this linear shape function. With a pretty random scaling factor I could get lines of the form `I = 256 - x / (excitation * factor)` to pass through all the points. So this means that each spot in the diffraction pattern has the same "base intensity", reduced linearly by the excitation error.

I vaguely remember someone telling me that ASTAR does not take into consideration atomic scattering factors, or at least the angular dependence of atomic scattering factors, which would be consistent with these results.

These are the images for the same crystal structure using diffsims

Using the linear shape factor:
![image](https://user-images.githubusercontent.com/25320842/115953543-680f4380-a4ec-11eb-9435-be76c9290b8d.png)

Using the Lorentzian shape factor:
![image](https://user-images.githubusercontent.com/25320842/115953532-57f76400-a4ec-11eb-8150-1a07084c98c4.png)

The difference between different excitation errors is (within a certain range) almost negligible. We don't see this linear relationship and the intensity is clearly dominated by the structure factor (which includes atomic scattering). Still I find it odd that there is almost no difference at all and will investigate whether this is a bug somewhere.

The consequence is that in Pyxem spots near the origin dominate in the correlation calculation. Spots further away tend to be more sensitive to changes in orientation, so boosting their intensity to "weigh" more could improve indexation fidelity.

[...]

Issue #159, opened 2021-04-06
Improving/bugfixing how we simulate precession


pc494:
I think our current precession simulations could be improved a little. Currently we use "max_excitation_error" to eliminate spots from our precessed simulations, but this isn't quiet accurate as it doesn't account for the size of the g vector sensibly. 

Case 1 (no approximation):
This finds the average excitation error, but a spot can have an average excitation error greater than max_excitation_error and we would still expect to see it in a pattern as long as somewhere on the circuit it had a sufficiently small excitation error. This is a bug (I think).

Case 2 (approximation):
These spots are eliminated before we even consider that there is precession going on.

Proposed solution:
-------------------------

Working from Equation 10 of [1] which provides the intensity of a spot as a function of:

g-vector (we have this)
precession angle (we have this)
the unprecessed excitation error, which is exactly what we have
the extinction distance (we can use max_excitation_error as a proxy)

(and is currently used in "lorenztian precession)

What Figure 3 (same manuscript) shows is that changing max_excitation_error doesn't change the width of the camel plot significantly. This is likely because we are in the regime (below equation 11) where we have full integration of our spots. We could create two cases (basically by hand) and use them as the eliminations. They would be:

~ numerical_factor * g * phi (precesison regime)
~ The 98th percentile (or similar) of the phi = 0 lorenztian 

I would take this opportunity to remove the "manual integration" as I'm not sure this is a particularly good option. 

[1] - Specifics of the data processing of precession electron diffraction tomography data and their implementation in the program PETS2.0 - Lukáš Palatinus et al.

Code
-------

As it currently is.

```python
# Identify points intersecting the Ewald sphere within maximum
        # excitation error and store the magnitude of their excitation error.
        r_sphere = 1 / wavelength
        r_spot = np.sqrt(np.sum(np.square(cartesian_coordinates[:, :2]), axis=1))
        z_spot = cartesian_coordinates[:, 2]

        if self.precession_angle > 0 and not self.approximate_precession:
            # We find the average excitation error - this step can be
            # quite expensive
            excitation_error = _average_excitation_error_precession(
                z_spot,
                r_spot,
                wavelength,
                self.precession_angle,
            )
        else:
            z_sphere = -np.sqrt(r_sphere ** 2 - r_spot ** 2) + r_sphere
            excitation_error = z_sphere - z_spot
        # Mask parameters corresponding to excited reflections.
        intersection = np.abs(excitation_error) < max_excitation_error
[...]

PR #166, opened 2021-05-04
Added ignore atomic scatter option + bugfixes precession

din14970:
**Release Notes**
* In `DiffractionGenerator`, `scattering_params` now takes `None` as an option, which acts to ignore atomic scattering altogether. This emulates the way ASTAR simulates patterns if the `linear` shape function is used.
* Changed the `Lorentzian` shape factor model such that the peak is always at 1.
* I added an extra parameter `shape_factor_width` in `calculate_ed_data` to optionally decouple the "cutting" operation from the shape factor width. If not set it will take the same value as `max_excitation_error`. For linear shape functions it does not make sense to have them be different, but for distributions with wide infinite tails it might (otherwise you "cut" out reflections always at the same value of the shape factor).
* I fixed the linear shape function so that it can not be negative
* When simulating with precession I added a pre-filter using the `max_excitation_error` so that the calculation does not have to be performed on the entire sphere of reciprocal space points. I make the simple assumption that points must lie between or on the cone extremes swept out by the precessing sphere (approximated as a plane). Since `max_excitation_error` and `shape_factor_width` are decoupled, any possible error (incorrectly removed reflections) from a very curved Ewald sphere should be correctable.
* Some docstring clarifications and fixes
* black formatting

See issues #159 and #163. I took another look at the ED simulation procedure and it looks pretty solid. The main point I added is the option to simulate without taking atomic scattering into consideration, which boosts the relative intensity of spots further away from the origin - this should emulate the behavior of how ASTAR simulates patterns. 

**Work in progress?**
- [x] Improve test coverage to consider `scattering_params = None` option
- [ ] Update changelog

#### For reviewers

  later.

### DIFF
--- a/diffsims/generators/diffraction_generator.py
+++ b/diffsims/generators/diffraction_generator.py
@@ -50,6 +50,7 @@ _shape_factor_model_mapping = {
     "lorentzian": lorentzian,
 }
 
+
 def _shape_factor_precession(
     excitation_error, r_spot, phi, shape_function, max_excitation, **kwargs
 ):
@@ -92,12 +93,13 @@ def _shape_factor_precession(
         def integrand(theta):
             # Equation 8 in L.Palatinus et al. Acta Cryst. (2019) B75, 512-522
             S_zero = excitation_error_i
-            variable_term = r_spot_i*(phi)*np.cos(theta)
+            variable_term = r_spot_i * (phi) * np.cos(theta)
             return shape_function(S_zero + variable_term, max_excitation, **kwargs)
 
         # average factor integrated over the full revolution of the beam
-        shf[i] = (1 / (2*np.pi)) * quad(integrand, 0, 2*np.pi)[0]
+        shf[i] = (1 / (2 * np.pi)) * quad(integrand, 0, 2 * np.pi)[0]
     return shf
+
 
 class DiffractionGenerator(object):
     """Computes electron diffraction patterns for a crystal structure.
@@ -118,9 +120,10 @@ class DiffractionGenerator(object):
     accelerating_voltage : float
         The accelerating voltage of the microscope in kV.
     scattering_params : str
-        "lobato" or "xtables"
+        "lobato", "xtables" or None. If None is provided then atomic
+        scattering is not taken into consideration.
     precession_angle : float
-        Angle about which the beam is precessed. Default is no precession.
+        Angle about which the beam is precessed in degrees. Default is no precession.
     shape_factor_model : function or string
         A function that takes excitation_error and
         `max_excitation_error` (and potentially kwargs) and returns
@@ -170,7 +173,7 @@ class DiffractionGenerator(object):
             self.shape_factor_model = shape_factor_model
         self.minimum_intensity = minimum_intensity
         self.shape_factor_kwargs = kwargs
-        if scattering_params in ["lobato", "xtables"]:
+        if scattering_params in ["lobato", "xtables", None]:
             self.scattering_params = scattering_params
         else:
             raise NotImplementedError(
@@ -186,6 +189,7 @@ class DiffractionGenerator(object):
         rotation=(0, 0, 0),
         with_direct_beam=True,
         max_excitation_error=1e-2,
+        shape_factor_width=None,
         debye_waller_factors={},
     ):
         """Calculates the Electron Diffraction data for a structure.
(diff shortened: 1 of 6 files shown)


din14970:
@pc494 I would tentatively say this is ready for review, then we can decide what goes into the changelog if you are happy

diffsims/generators/diffraction_generator.py
```diff
@@ -244,26 +253,36 @@ def calculate_ed_data(
...
-            intersection_coordinates = cartesian_coordinates[intersection]
-            excitation_error = excitation_error[intersection]
-            r_spot = r_spot[intersection]
-            g_indices = g_indices[intersection]
-            g_hkls = g_hkls[intersection]
-
+        else:
+            # only consider points that are between or touch the cones created by the ewald sphere extremes
+            # ewald sphere is approximated flat
+            intersection = np.abs(z_spot) - np.abs(
```
pc494: I'm nervous about introducing a flat ewald sphere when we don't have to; is the speed loss is crippling?

diffsims/utils/shape_factor_models.py
```diff
@@ -38,6 +38,7 @@ def binary(excitation_error, max_excitation_error):
     return 1
 
 
+@np.vectorize
```
pc494, line 41: what does this do?

diffsims/utils/sim_utils.py
```diff
@@ -167,16 +169,20 @@ def get_vectorized_list_for_atomic_scattering_factors(
...
     n_structures = len(structure)
     coeffs = np.empty((n_structures, 5, 2))
     fcoords = np.empty((n_structures, 3))
     occus = np.empty(n_structures)
     dwfactors = np.empty(n_structures)
+    default = np.zeros((5, 2))
 
     for i, site in enumerate(structure):
-        coeffs[i] = scattering_params_dict[site.element]
+        coeffs[i] = scattering_params_dict.get(site.element, default)
```
pc494, line 185: What are these changes about?

pc494 (commented):
This review just has a couple of questions in it.

diffsims/utils/shape_factor_models.py
din14970: makes a scalar function applicable onto an array on an element-wise basis. I found this the easiest way to implement this top-hat function, where if the result is less than 0 it is set to 0. Previously the linear shape factor could be negative, which in my view was actually a bug.

diffsims/utils/sim_utils.py
din14970: This is to deal with the case where `None` is passed as scattering params. To avoid having to write separate cases with lots of if-else statements, if the scattering params dict is empty or does not contain the element it just puts in 0's for the coefficients.  This is to avoid having to branch more than necessary in `get_kinematical_intensities`

pc494: Yes, sorry this was me being slow, that's clever

diffsims/utils/shape_factor_models.py
pc494: Does that mean we get vectorized calculation over `excitation_errors` in `.calculate_ed_data`?

diffsims/generators/diffraction_generator.py
din14970: For very simple structures no, but once you have thousands of reciprocal space points for more complex structures (which I had) it makes a big difference. If the ewald sphere is not flat and one is afraid of cutting out too many reflections, max_excitation_error can just be increased and one can use shape_factor_width independently to control spot intensity. Spots that are incorrectly still included may be removed then using the min_intensity. So in my view there is no issue of either missing spots or additional spots. But I'll check whether a curved Ewald sphere can be taken into consideration nonetheless, shouldn't be too much more complicated.

diffsims/utils/shape_factor_models.py
din14970: yes exactly, you can pass in an array of excitation errors into this function. if `np.vectorized` was not there, it would get hung up on the inequality in the function to set negatives to 0.

diffsims/generators/diffraction_generator.py
din14970: I've now updated the filter to take into consideration the ewald sphere curvature

pc494:
Great, I think all we need now is the changelog update, with 

added: option to use `None` for scattering_params

pc494:
Merging

PR #167, opened 2021-05-13
Minor bugfixes

din14970:
---
name: Minor bugfixes
about: Added a kwarg and fixed linear shape function

---

**Release Notes**
The new keyword argument shape factor width was not implemented in the diffraction generator. This is now done so that it can be used when generating a diffraction library.

In addition, I discovered as strange bug that I couldn't pickle the diffraction library when I used the `linear` shape function to calculate it. I traced this back to my use of `np.vectorize` on the `linear` shape function, so I had to alter this function to get it to work. Now it looks less elegant in my opinion because I have to consider the case whereby an array is passed in and when a simple number is passed in (occurs when doing the full precession calculation). If there is a more elegant way to do this let me know.

#### For reviewers

  later.

### DIFF
--- a/diffsims/generators/library_generator.py
+++ b/diffsims/generators/library_generator.py
@@ -52,6 +52,7 @@ class DiffractionLibraryGenerator:
         half_shape,
         with_direct_beam=True,
         max_excitation_error=1e-2,
+        shape_factor_width=None,
         debye_waller_factors={},
     ):
         """Calculates a dictionary of diffraction data for a library of crystal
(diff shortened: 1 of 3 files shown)


pc494:
Am I right in thinking that this is ready for review @din14970?

din14970:
@pc494 Yes, although I would also suggest a small edit to the `crystal_system_dictionary` for the hexagonal system which I've been dealing with recently. In my view, the coordinates for the coordinates of the fundamental zone should be `[(0, 0, 0, 1), (1, 0, -1, 0), (2, -1, -1, 0)]`, at least when making a templates grid. Not sure if you would have this in this PR or another one.

pc494:
> @pc494 Yes, although I would also suggest a small edit to the `crystal_system_dictionary` for the hexagonal system which I've been dealing with recently. In my view, the coordinates for the coordinates of the fundamental zone should be `[(0, 0, 0, 1), (1, 0, -1, 0), (2, -1, -1, 0)]`, at least when making a templates grid. Not sure if you would have this in this PR or another one.

Do you mean the streotriangle?  I'll have a look

hakonanes:
The plane normals suggested by @din14970 seem reasonable:
m1
 m2

Issue #170, opened 2021-06-07
Better Defining Diffsims Purpose


CSSFrancis:
I realize that this feature request is a pretty vague but I wanted to just bring up some changes to diffsims that I am interested in making, most of which surrounding adding some more tools for visualization and labeling of simulations.  

I think diffsims preforms simulations very well, but it struggles when you have to visualize these simulations or put them into context (partially because originally it was never really meant for visualization)

**A couple of examples where some additional information might be helpful:**

- Electron spot patterns including diffraction spot labels
- Electron profile data including labels for peaks 
- Diffraction Simulations returning `DiffractionSignals` with proper metadata scales etc. @pc494 what is your opinion on adding in a pyxem dependancy?
- Zone Axis Patterns returned with better navigation

In general I think that diffsims could be particularly useful as a substitution for something like JEMS.  Simulating structures from CIF files, showing zone axis patterns etc.. but that would probably require leaning on pyxem/hyperspy pretty heavily for visualization and labeling. (which I think was originally something that diffsims wanted to avoid) 

There is also the option to do this within the scope of diffsims (and not add pyxem as a dependancy) but I feel like that would be a lot of recreating many of the features already in pyxem/hyperspy.

@din14970  and @hakonanes  you might have insights into this as well.

din14970:
Thanks for pinging me for the discussion @CSSFrancis 

For me very simply the scope of diffsims is to easily and flexibly calculate kinematical diffraction intensities. An expanded scope may be including dynamical diffraction in the calculation. I don't think multislice simulations for e.g. CBED patterns are in the scope, but I'm open to the discussion. Overall, the output of the simulation is a list of coordinates in reciprocal space (1D = profile, 2D = ED pattern), their miller coordinates, and a list of associated intensities.

When I started working on diffsims I was also missing this quick and dirty visualization functionality and added the `plot` method (I first added it as a function and @pc494 moved it to a more appropriate location) https://github.com/pyxem/diffsims/blob/b239a7e83ee38751fdc58c9b9797ab58aeef4567/diffsims/sims/diffraction_simulation.py#L217, so perhaps you can work from this. For a couple figures I made, I also wanted the option for labels to create something like this:
![image](https://user-images.githubusercontent.com/25320842/121142222-11f82480-c83c-11eb-9345-eb90abfd1a38.png)

> Electron spot patterns including diffraction spot labels

[...]

Issue #168, opened 2021-05-26
Select more natural corners for the streographic triangle (hexagonal system)


pc494:
From #167:

"I would also suggest a small edit to the crystal_system_dictionary for the hexagonal system which I've been dealing with recently. In my view, the coordinates for the coordinates of the fundamental zone should be [(0, 0, 0, 1), (1, 0, -1, 0), (2, -1, -1, 0)], at least when making a templates grid."

> The plane normals suggested by @din14970 seem reasonable:
> ![m1](https://user-images.githubusercontent.com/12139781/119660539-aa9ba700-be2f-11eb-95d2-c185da4fe58e.png)
> ![m2](https://user-images.githubusercontent.com/12139781/119660541-ab343d80-be2f-11eb-82a6-61ffa46fb77c.png)

hakonanes:
Good. I ment crystal directions in that comment...

I assume that dictionary can be replaced by `SphericalRegion.vertices` which I hope to implement in orix soon, https://github.com/pyxem/orix/issues/169? If there is any interest in that, I can try to get the code I have locally finished and make a PR to orix soon.

din14970:
@hakonanes I think if we can get rid of the sketchy dictionary in diffsims and replace it with something a bit more robust it's probably good. However, something that must be kept clear is what the zone really represents. If it is the "set of unique beam directions" as used in an IPF color key, this is still something else than the set of unique orientations with the constraint `phi_1 = 0`, because if the crystal has inversion symmetry some identical beam directions can still produce orientations that are not related to each other by a `phi_1` rotation but a rotation and an inversion. In the example of template matching, this manifests as mirrored templates. For generating the template library, I still only use the "fundamental" beam-directions orientations to conserve memory and time calculating the templates, but during the indexation step I also mirror the templates and match those as well. While the difference is very slight, there is always a clear preference:
<img width="262" alt="image" src="https://user-images.githubusercontent.com/25320842/119959617-73023b80-bfa4-11eb-8b20-1b52fdb94dd7.png">
What is therefore very important is that the Euler angles corresponding to the templates  in the mirrored sector are correct: they should match the templates mirrored across the x axis (to calculate them I basically reverse theta). For this to be the case, in the current implementation my fundamental zone should always have the corners `[001]`, a vector `[x01]` with x>0 and `[yz1]` with z>0 and no constraint on y.

pc494:
This code is now orix's problem

PR #172, opened 2021-07-06
Label diffraction spots in plot

din14970:
---
name: Option to add labels in simulation diffraction spots
---

**Release Notes**
> extra parameters in diffraction pattern's `plot` method for drawing miller index labels next to the diffraction spots

After discussion in #170 I thought I would open this small PR for adding the option to draw the miller indices of spot patterns. With this change it is now possible to do the following:

```python
fig, ax = plt.subplots()
ax.set_aspect("equal")
ax.set_xlim(-2, 2)
ax.set_ylim(2, -2)
simtest.plot(show_labels=True, label_offset=(0, -0.2), 
             label_formatting={"fontsize": 8},
             size_factor=10, ax = ax)
```

afbeelding

Does this approach what you had in mind @CSSFrancis ?

Apparently there are also two small additions: the option to mirror pattern masks and the update to the stereo triangle for the hex system. It seems this was not merged before, see #168 

I would have liked to create a vectorized implementation but have no idea if this is possible with strings. Probably one could do some smart tricks with string formatting, but I couldn't figure out how to do that with the `\bar{}` on negative indices. If anyone has a better implementation idea feel free to modify.

**Work in progress?**
- [ ] Needs a test to be added for coverage

#### For reviewers

  later.

### DIFF
--- a/diffsims/generators/rotation_list_generators.py
+++ b/diffsims/generators/rotation_list_generators.py
@@ -43,7 +43,7 @@ from diffsims.generators.sphere_mesh_generators import (
 # https://github.com/pyxem/orix/issues/125#issuecomment-698956290.
 crystal_system_dictionary = {
     "cubic": [(0, 0, 1), (1, 1, 1), (1, 0, 1)],
-    "hexagonal": [(0, 0, 0, 1), (1, 0, -1, 0), (-1, 2, -1, 0)],
+    "hexagonal": [(0, 0, 0, 1), (1, 0, -1, 0), (2, -1, -1, 0)],
     "trigonal": [(0, 0, 0, 1), (-2, 1, 1, 0), (-1, 2, -1, 0)],
     "tetragonal": [(0, 0, 1), (1, 0, 0), (1, 1, 0)],
     "orthorhombic": [(0, 0, 1), (-1, 0, 0), (0, 1, 0)],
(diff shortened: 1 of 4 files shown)


hakonanes:
This looks like an important "pedagogical" feature!

> I would have liked to create a vectorized implementation but have no idea if this is possible with strings. Probably one could do some smart tricks with string formatting

I don't think matplotlib.pyplot.text supports passing more than one set of position and text. Is that what you had in mind? If you find a way to do that, I would be interested, because we have labeling of zone axes in EBSD patterns in kikuchipy ([user guide](https://kikuchipy.org/en/latest/geometrical_ebsd_simulations.html), [API](https://kikuchipy.org/en/latest/reference.html#kikuchipy.simulations.GeometricalEBSDSimulation.zone_axes_labels_as_markers), and [zone_axes_labels_as_markers()](https://github.com/pyxem/kikuchipy/blob/master/kikuchipy/simulations/geometrical_ebsd_simulation.py#L251-L283) which calls [get_text_list()](https://github.com/pyxem/kikuchipy/blob/07fa25f93e9052725592d23f95ae639f85d94a3f/kikuchipy/draw/markers.py#L84)).

din14970:
@pc494 @hakonanes I think with this minor edit it would be fine for review now

pc494 (approved):
LGTM

PR #175, opened 2021-09-20
Improve DiffractionSimulation

### LINKED ISSUE #174: Weird behavior DiffractionSimulation.intensities
**Describe the bug**
Due to the way `intensities` is implemented as a property with a setter, some weird behavior can occur due to the direct beam mask.

Similar issues can be created with the `coordinates` attribute.

**To Reproduce**
```python
from diffsims.sims.diffraction_simulation import DiffractionSimulation

coords = np.array([[0, 0], [1, 1], [2, 3]])
intensities = np.array([1, 1, 3])
sim = DiffractionSimulation(coordinates = coords, intensities = intensities, with_direct_beam=False)
sim.intensities = sim.intensities

# now throws an error because intensities is of the wrong shape
sim.intensities
```

**Expected behavior**
The setter needs to be more restrictive so that not just any array can be used, which creates inconsistencies.
I propose to check the size of the passed arrays and modify behavior depending on whether `with_direct_beam` is on or not.

din14970:
---
name: Fix DiffractionSimulation
about: see #174

---

**Release Notes**
Reworked the way `DiffractionSimulation` behaves, thereby fixing issues setting properties like `coordinates`, `indices` and `intensities`. 

This PR aims to fix #174 but also reworks the class a bit

The changes mainly pertain to the property/setter behavior of the `coordinates`, `indices` and `intensities` attributes. Previously `DiffractionSimulation` could exist as an empty shell without any diffraction spots, and coordinates, indices and intensities could be set to anything. This can cause all kinds of issues with the getters, that relied on `_coordinates`, etc. to be either all arrays of the same size or `None`. But this was not enforced anywhere, plus one ran into trouble with the `with_direct_beam` masking.

I have changed `coordinates` to a mandatory argument in the `__init__`.  My reasoning is that a diffraction pattern without any spots doesn't make much sense. A completely empty array can still be passed. Indices and intensities can be left empty but appropriately sized arrays of zeros will be automatically created based on the size of `coordinates`. In the `init` method, the shape of all these arrays is verified, and `_coordinates`, `_indices` and `_intensities` are set based on what is passed in. The getters and setters of `indices` `intensities` and `coordinates` can then be substantially simplified, both rely on `self.direct_beam_mask` and stay consistent in shape amongst each other since I use numpy slicing to update the values. Changing the number of spots is no longer allowed, for this one should simply create a new `DiffractionSimulation`. What can be done instead now is slicing into the pattern, for example

```
sim[sim.intensities > 8]
```

in order to filter out spots based on some condition. This makes sense as long as `coordinates`, `indices`, and `intensities` are the same size. **note that on a slice like this, information about the direct beam is lost if it was already "filtered out" by `with_direct_beam`**

Other changes I've made were mainly as a consequence of the previous changes and also make sense in my opinion. `calibrated_coordinates` only makes sense in the x-y plane, there is no reason it should return the z-coordinate since there is no calibration value for it. One might consider renaming this `detector_coordinates`?

### 30/09/21
I have tried to clean up this class a bit more and add a couple more features:
- allow for "adding" and "extending" patterns with other patterns, effectively concatenating coordinate, index and intensity arrays. Basically
```
sim1 + sim2
```
Will produce another `DiffractionSimulation` object with the combined spots of the original two. With the `extend` method, spots can be added from another pattern in-place, just like it exists in ASE.
- added rotation, mirroring and translation arguments for plotting, creating a mask and creating a gaussian image of the pattern. It is convenient to be able to rotate, flip and shift the templates easily. I also added a method to rotate, shift, mirror the template spot coordinates in-place.
- The method for creating a diffraction pattern image was reworked to allow for non-square images
- After the issue in #176 , I think it is unreasonable to set calibration to 1 by default. Instead I set it to none by default and whenever a user tries to access `calibrated_coordinates` or a method that depends on it, an exception is raised telling the user the calibration should be set.

**Work in progress?**
- [x] add testing for __getitem__
- [ ] update changelog

#### For reviewers

  later.

### DIFF
--- a/diffsims/sims/diffraction_simulation.py
+++ b/diffsims/sims/diffraction_simulation.py
@@ -20,9 +20,10 @@ import matplotlib.pyplot as plt
 import numpy as np
 from diffsims.pattern.detector_functions import add_shot_and_point_spread
 from diffsims.utils import mask_utils
+from collections import Sequence
 
 
-class DiffractionSimulation:
+class DiffractionSimulation(Sequence):
     """Holds the result of a kinematic diffraction pattern simulation.
 
     Parameters
@@ -68,6 +69,21 @@ class DiffractionSimulation:
         self.calibration = calibration
         self.offset = offset
         self.with_direct_beam = with_direct_beam
+
+    def __len__(self):
+        return self.direct_beam_mask.shape[0]
+
+    def __getitem__(self, sliced):
+        coords = self.coordinates[sliced]
+        inds = self.indices[sliced]
+        ints = self.intensities[sliced]
+        return DiffractionSimulation(coords,
+                                     indices=inds,
+                                     intensities=ints,
+                                     calibration=self.calibration,
+                                     offset=self.offset,
+                                     with_direct_beam=self.with_direct_beam
+                                    )
 
     @property
     def indices(self):
--- a/diffsims/sims/diffraction_simulation.py
+++ b/diffsims/sims/diffraction_simulation.py
@@ -44,7 +44,7 @@ class DiffractionSimulation:
 
     def __init__(
         self,
-        coordinates=None,
+        coordinates,
         indices=None,
         intensities=None,
         calibration=1.0,
@@ -54,25 +54,33 @@ class DiffractionSimulation:
         """Initializes the DiffractionSimulation object with data values for
         the coordinates, indices, intensities, calibration and offset.
         """
-        self._coordinates = None
-        self.coordinates = coordinates
-        self.indices = indices
-        self._intensities = None
-        self.intensities = intensities
-        self._calibration = (1.0, 1.0)
+        if indices is None:
+            indices = np.zeros((coordinates.shape[0], 3))
+        if intensities is None:
+            intensities = np.zeros((coordinates.shape[0]))
+        # check here whether shapes are all the same
+        if coordinates.shape[0] == indices.shape[0] == intensities.shape[0]:
+            self._coordinates = coordinates
+            self._indices = indices
+            self._intensities = intensities
+        else:
+            raise ValueError("Coordinate, intensity, and indices lists must be same size.")
         self.calibration = calibration
         self.offset = offset
         self.with_direct_beam = with_direct_beam
 
     @property
+    def indices(self):
+        return self._indices[self.direct_beam_mask]
+
+    @indices.setter
+    def indices(self, indices):
+        self._indices[self.direct_beam_mask] = indices
+
+    @property
     def calibrated_coordinates(self):
         """ndarray : Coordinates converted into pixel space."""
-        coordinates = np.copy(self.coordinates)
-        coordinates[:, 0] += self.offset[0]
-        coordinates[:, 1] += self.offset[1]
-        coordinates[:, 0] /= self.calibration[0]
-        coordinates[:, 1] /= self.calibration[1]
-        return coordinates
+        return (self.coordinates[:, :2] + np.array(self.offset))/np.array(self.calibration)
 
     @property
     def calibration(self):
@@ -106,24 +114,20 @@ class DiffractionSimulation:
     @property
     def coordinates(self):
         """ndarray : The coordinates of all unmasked points."""
-        if self._coordinates is None:
-            return None
         return self._coordinates[self.direct_beam_mask]
 
     @coordinates.setter
     def coordinates(self, coordinates):
-        self._coordinates = coordinates
+        self._coordinates[self.direct_beam_mask] = coordinates
 
     @property
     def intensities(self):
         """ndarray : The intensities of all unmasked points."""
-        if self._intensities is None:
-            return None
         return self._intensities[self.direct_beam_mask]
 
     @intensities.setter
     def intensities(self, intensities):
-        self._intensities = intensities
+        self._intensities[self.direct_beam_mask] = intensities
 
     def get_as_mask(self, shape, radius=6., negative=True,
                     radius_function=None, direct_beam_position=None,
@@ -160,7 +164,7 @@ class DiffractionSimulation:
         cx, cy = shape[0]//2, shape[1]//2
         if direct_beam_position is not None:
             cx, cy = direct_beam_position
-        point_coordinates_shifted = self.calibrated_coordinates[:, :-1].copy()
+        point_coordinates_shifted = self.calibrated_coordinates.copy()
         x = point_coordinates_shifted[:, 0]
         y = point_coordinates_shifted[:, 1]
         mirrored_factor = -1 if mirrored else 1
@@ -269,11 +273,6 @@ class DiffractionSimulation:
 
         if show_labels:
             millers = self.indices.astype(np.int16)
-            if millers.shape[0] != coords.shape[0]:
-                # ensure we don't label 0,0,0, smallest miller index is 001
-                miller_dist = np.linalg.norm(millers, axis=1)
-                condition = miller_dist > 0.1
-                millers = millers[condition]
             # only label the points inside the axes
             xlim = ax.get_xlim()
             ylim = ax.get_ylim()
--- a/diffsims/sims/diffraction_simulation.py
+++ b/diffsims/sims/diffraction_simulation.py
@@ -20,10 +20,10 @@ import matplotlib.pyplot as plt
 import numpy as np
 from diffsims.pattern.detector_functions import add_shot_and_point_spread
 from diffsims.utils import mask_utils
-from collections import Sequence
+import copy
 
 
-class DiffractionSimulation(Sequence):
+class DiffractionSimulation:
     """Holds the result of a kinematic diffraction pattern simulation.
 
     Parameters
@@ -75,13 +75,19 @@ class DiffractionSimulation(Sequence):
                 "Coordinate, intensity, and indices lists must be of the correct and matching shape."
             )
         self.calibration = calibration
-        self.offset = offset
+        self.offset = np.array(offset)
         self.with_direct_beam = with_direct_beam
 
     def __len__(self):
         return self.coordinates.shape[0]
 
+    @property
+    def size(self):
+        return self.__len__()
+
     def __getitem__(self, sliced):
+        """Sliced is any valid numpy slice that does not change the number of
+        dimensions or number of columns"""
         coords = self.coordinates[sliced]
         inds = self.indices[sliced]
         ints = self.intensities[sliced]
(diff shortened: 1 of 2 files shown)


din14970:
Coveralls is not accessible at the moment which seems to be the reason why CI is failing

hakonanes:
FYI, Coveralls [now works](https://status.coveralls.io/?utm_source=embed)!

*Edit: At least the workflows don't fail. See [the reduced coverage](https://github.com/pyxem/diffsims/pull/175/checks?check_run_id=3665738260#step:9:78).*

din14970:
Think this should be ok for review, if you agree with the features I've added I'll update the changelog

hakonanes:
I'll have a look tomorrow, but since I haven't used this part of diffsims, someone else who has should have a look as well before merging.

diffsims/sims/diffraction_simulation.py
```diff
@@ -20,9 +20,10 @@
 import numpy as np
 from diffsims.pattern.detector_functions import add_shot_and_point_spread
 from diffsims.utils import mask_utils
+from collections import Sequence
 
 
-class DiffractionSimulation:
+class DiffractionSimulation(Sequence):
```
hakonanes: Are there any of the `Sequence` class methods or properties inherited that are useful for this class? If not, I would remove the inheritance.

diffsims/sims/diffraction_simulation.py
```diff
@@ -44,35 +45,103 @@ class DiffractionSimulation:
...
+            indices = np.full((coordinates.shape[0], 3), np.nan)
+        if intensities is None:
+            intensities = np.full((coordinates.shape[0]), np.nan)
+        # check here whether shapes are all the same
+        if (
+            coordinates.shape[0] == indices.shape[0] == intensities.shape[0]
+            and coordinates.ndim == indices.ndim == 2
+            and intensities.ndim == 1
+        ):
+            self._coordinates = coordinates
```
hakonanes, line 70: Ideally, you should use the `coordinates.setter` here, without having to set `_coordinates`. The shape checks could then be part of indices/intensities setters. At least `self.indices = indices` can be used here, since `self.direct_beam_mask` does not use `self._indices`, right?

diffsims/sims/diffraction_simulation.py
```diff
@@ -44,35 +45,103 @@ class DiffractionSimulation:
...
+            self._intensities = intensities
+        else:
+            raise ValueError(
+                "Coordinate, intensity, and indices lists must be of the correct and matching shape."
+            )
         self.calibration = calibration
         self.offset = offset
         self.with_direct_beam = with_direct_beam
 
+    def __len__(self):
```
hakonanes, line 81: Is `len(diff_sim)` the most natural way to get number of coordinates? I usually add a `size` attribute if it provides a much used quantity of that class, like number of reciprocal lattice points in a `ReciprocalLatticePoint` instance. As a user I would look for such an attribute before thinking of using `len()`, but that might not be the case for everyone.

diffsims/sims/diffraction_simulation.py
```diff
@@ -44,35 +45,103 @@ class DiffractionSimulation:
...
+        return self.coordinates.shape[0]
+
+    def __getitem__(self, sliced):
```
hakonanes, line 88: Would be nice with a very short docstring explaining that `sliced` must be a NumPy array of a certain shape.

diffsims/sims/diffraction_simulation.py
```diff
@@ -44,35 +45,103 @@ class DiffractionSimulation:
...
+            offset=self.offset,
+            with_direct_beam=self.with_direct_beam,
+        )
+
+    def extend(self, other):
+        """Add the diffraction spots from another DiffractionSimulation"""
+        coords, inds, ints = self._combine_data(other)
+        self._coordinates = coords
+        self._indices = inds
+        self._ints = ints
```
hakonanes: `add()` returns a new combined instance, while `extend()` combines inplace? Could you achieve the same thing by providing a `deepcopy()` method (one-line method using standard library `copy.deepcopy(self)`), and just keep the `extend()` method?

Would `extend()` be more robust if you use the `setter`s directly, and add some checks to them, or to `_combine_data()`? It looks like these two methods can easily make NumPy raise errors. Perhaps adopt the checks from `__init__()`?

diffsims/sims/diffraction_simulation.py
hakonanes: This goes for other methods and properties which aren't documented in the class docstring as well.

diffsims/sims/diffraction_simulation.py
```diff
@@ -44,35 +45,103 @@ class DiffractionSimulation:
...
     def calibrated_coordinates(self):
         """ndarray : Coordinates converted into pixel space."""
-        coordinates = np.copy(self.coordinates)
-        coordinates[:, 0] += self.offset[0]
-        coordinates[:, 1] += self.offset[1]
-        coordinates[:, 0] /= self.calibration[0]
-        coordinates[:, 1] /= self.calibration[1]
-        return coordinates
+        if self.calibration is not None:
+            return (self.coordinates[:, :2] + np.array(self.offset)) / np.array(
```
hakonanes: Is there any reason for `offset` to not be a NumPy array instead of a tuple by default?

diffsims/sims/diffraction_simulation.py
```diff
@@ -82,6 +151,9 @@ def calibration(self):
 
     @calibration.setter
     def calibration(self, calibration):
+        if calibration is None:
+            self._calibration = calibration
+            return
```
hakonanes: Can we avoid the `return`? Perhaps just change `calibration` under any `True` condition, and then just set it at the end?

```python
if calibration is None:
    pass
elif np.allclose(calibration, 0):
   ...
elif isinstance():
   ...
elif len():
   ...
else:
   ...
self._calibration = calibration
```

diffsims/sims/diffraction_simulation.py
```diff
@@ -106,29 +178,77 @@ def direct_beam_mask(self):
...
+        y = coords_transformed[:, 1]
+        mirrored_factor = -1 if mirrored else 1
+        theta = mirrored_factor * np.arctan2(y, x) + np.deg2rad(angle)
+        rd = np.sqrt(x ** 2 + y ** 2)
+        coords_transformed[:, 0] = rd * np.cos(theta) + cx
+        coords_transformed[:, 1] = rd * np.sin(theta) + cy
+        return coords_transformed
+
+    def rotate_shift_coordinates(self, angle, center=(0, 0), mirrored=False):
+        """Translate, rotate or mirror the pattern"""
```
hakonanes: The docstring must be added here to the public API method, not the private one. The `units` can be documented in the private method docstring, pointing to the public API docstring.

diffsims/tests/sims/test_diffraction_simulation.py
```diff
@@ -88,13 +88,93 @@ def test_plot_profile_simulation(profile_simulation):
...
+    def test_single_spot(self):
+        assert DiffractionSimulation(np.array([1, 2, 3])).coordinates.shape == (1, 3)
+
+    def test_get_item(self, diffraction_simulation):
+        assert diffraction_simulation[1].coordinates.shape == (1, 3)
+        assert diffraction_simulation[0:2].coordinates.shape == (2, 3)
+
+    @pytest.mark.xfail(raises=ValueError)
+    def test_fail_get_item(self, diffraction_simulation):
+        diffraction_simulation[None]
```
hakonanes: I think the xfail fields clutter the pytest output: I would instead use

```python
def test_fail_get_item(self, diffraction_simulation):
    with pytest.raises(ValueError, match="part or the full expected error message"):
        _ = diffraction_simulation[None]
```

hakonanes (commented):
Looks like solid additions! I've made general comments and suggestions on the structure of added functionality, which I believe could make it less error prone and easier to maintainable.

(Would be great to have a user guide notebook or something to see the added functionality in action, but that is for another time, and other people perhaps.)

diffsims/sims/diffraction_simulation.py
din14970: I saw an example online where they used this construct when defining a custom `__getitem__`. But it seems it's not necessary, so I will remove it.

diffsims/sims/diffraction_simulation.py
din14970: I added len because it was a necessary method for subclassing `Sequence`. While it is now not strictly necessary anymore I might keep it, but I would also implement `size` as a property as you suggest.

diffsims/sims/diffraction_simulation.py
din14970: Just for reference any numpy slice notation is valid, e.g. also `2:5`. However, numpy is very generous in the allowed slicing, for example passing in `None` can create new axes which would break the new object. I only allow slices that keep the array dimensions the same.

diffsims/sims/diffraction_simulation.py
din14970: Here I disagree. This is kind of how it was implemented before and was giving the problems. The `_` values contain information for all the spots, including possibly the direct beam and stays a fixed size for the entire object life. There is no way to change them except by accessing and overwriting these private properties. The size of the `@property` values can vary by 1 depending on whether `self.with_direct_beam` is True or False, and this is the case for coordinates, indices and intensities. When one tries to set the values through the setter in the init, the logic becomes much more complicated and kind of circular due to this `with_direct_beam` mask thing, which itself relies on the size of the `_coordinates` array. The way I see it is the hidden `_` values should be used when I want to interact with all the reflection information stored in the object. I want to deal with the `@property` values when I want to get the values that the user sees (i.e. with or without direct beam).

din14970: the approach I have taken is to try and ensure a valid object is create in the init by doing a one time check on shapes and sizes of the passed in arrays. There should be no way to break the object after that unless a user explicitly changes the `_` attributes. As you see, setters now are not actually changing the references to new objects, they are changing array values in-place. This means that setters will automatically check whether the provided arrays are of the right size and this stays consistent. If `with_direct_beam` is True then one more value must be provided for all the array setters than if it is False. In my mind this makes sense, as these are the reflections that the user sees.

diffsims/sims/diffraction_simulation.py
din14970: > Could you achieve the same thing by providing a deepcopy() method (one-line method using standard library copy.deepcopy(self)), and just keep the extend() method?

Good idea

> Would extend() be more robust if you use the setters directly, and add some checks to them, or to _combine_data()? It looks like these two methods can easily make NumPy raise errors. Perhaps adopt the checks from __init__()?

As mentioned above, the main difference between the `_` attributes and the `@property` values is that with the latter you are not sure about the sizes.
The reason I use the `_` attributes is because otherwise an extend or addition operation will remove the direct beam if it is there in the original array and `with_direct_beam` is `False`. In my opinion, if one adds information, one should not throw away data, even if it is hidden. If the user changes the `with_direct_beam` property afterwards, the direct beam will be there.

diffsims/sims/diffraction_simulation.py
din14970: I think most people pass in arguments as tuples or lists and expect it to work. Since offset is only used in this method, I think explicitly converting to an array to allow broadcasting makes sense here. But I might move it to the init method.

diffsims/sims/diffraction_simulation.py
din14970: Good idea

diffsims/sims/diffraction_simulation.py
din14970: I also applied this to the calibration, it is automatically converted to a numpy array.

diffsims/sims/diffraction_simulation.py
hakonanes: This makes sense, keep it as it is.

I'm not saying get rid of the private attributes, these should be used. But I think it is good practice in general to have some checks with understandable error messages within the setters. But I wouldn't anything about that here now.

diffsims/sims/diffraction_simulation.py
hakonanes: I agree, data should be kept.

hakonanes (approved):
Looks good! I'll leave it up to @pc494 or @dnjohnstone to merge.

pc494:
Thanks for the contribution @din14970 and the review @hakonanes, apologies for my slowness.

Issue #198, opened 2023-10-17
Redesign of Simulations


CSSFrancis:
**Is your feature request related to a problem? Please describe.**
Simulating a Library of Diffraction patterns is (much) slower than it should be.  It also could be streamlined to allow for
ASE to be used as well as diffpy.

**Describe the solution you'd like**
I think there are a couple of things we should do in response to this:

1. Rewrite the Structure class as a `diffsims` class that extends the functionality of diffpy and ASE?
```python
class Structure:
    @cache
    def reciprocal_space_lattice(max_radius=10):
        pass
    @property
    def lattice():
        pass
    def plot()
        # visualize the atoms using ASE?
        pass
```

2. Clean up creating orientation lists with new `ConstrainedRotation` class defined by @din14970 

3. Add in some visualization tools to make things a bit more explicit

hakonanes:
For 1, I suggest to use orix' Phase class. Or extend it. It can also be changed to accommodate new needs. The Phase class was made to combine a Symmetry from orix and a Structure from diffpy.structure.

Note that diffsims also has the ReciprocalLatticeVector class for handling reflectors. This might be of some use?

CSSFrancis:
> For 1, I suggest to use orix' Phase class. Or extend it. It can also be changed to accommodate new needs. The Phase class was made to combine a Symmetry from orix and a Structure from diffpy.structure.

@hakonanes The `Phase` class looks nice!  I am a little curious is there is a way to have the `point_group`/ `space_group` initialized from the structure.  It seems like [spglib](https://spglib.readthedocs.io/en/latest/python-spglib.html) might be able to do this.  At least this is how ASE [handles](https://wiki.fysik.dtu.dk/ase/ase/spacegroup/spacegroup.html#ase.spacegroup.get_spacegroup) it. 

I like the idea of starting with some list of Phases and returning a list of `Phases` for some `CrystalMap` object. It would be nice to abstract out the use of `diffpy` as I think it is confusing (if only because the name is very similar to diffsims)
> 
> Note that diffsims also has the ReciprocalLatticeVector class for handling reflectors. This might be of some use?

Yea I probably need to try and tie some of these things together.  I think the currently `diffsims` has a lot of separate parts but not a great coherent structure so it would be good to start by defining what object we are trying to simulate and then go from there.

CSSFrancis:
Something like this might work:

[...]

PR #199, opened 2023-10-20
Add Deprecations

CSSFrancis:
Adds in Deprecation wrappers.

See Kikchipy, pyxem, hyperspy etc.

#### Progress of the PR

```python
>>> from diffsims.utils._deprecated import deprecate
>>> @deprecate(since=0.8, removal=0.9, alternative="bar")
>>> def foo(self, n):
>>>     return n + 1

>>> @property
>>> @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
>>> def this_property(self):
>>>     return 2
```

#### For reviewers

      in `.zenodo.json`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,7 @@ Added
 - Explicit support for Python 3.11.
 - Added deprecation tools for deprecating functions and arguments.
 - Added Pre-Commit for code formatting.
-
+- Added deprecation tools for deprecating functions, parameters, methods, and properties.
 
 Changed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Unreleased
 Added
 -----
 - Explicit support for Python 3.11.
+- Added deprecation tools for deprecating functions and arguments.
 - Added Pre-Commit for code formatting.
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,6 @@ Unreleased
 Added
 -----
 - Explicit support for Python 3.11.
-- Added deprecation tools for deprecating functions and arguments.
 - Added Pre-Commit for code formatting.
 - Added deprecation tools for deprecating functions, parameters, methods, and properties.
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -137,18 +137,16 @@ are raised whenever possible and feasible for functions/methods/properties/argum
 so that users get a heads-up one (minor) release before something is removed or changes,
 with a possible alternative to be used.
 
-The decorator should be placed right above the object signature to be deprecated::
+
+A deprecation decorator should be placed right above the object signature to be deprecated.
 
 .. code-block:: python
-    from diffsims.utils._deprecated import deprecate
-    @deprecate(since=0.8, removal=0.9, alternative="bar")
+    from diffsims.utils._deprecated import deprecated
+    @deprecated(since=0.8, removal=0.9, alternative="bar")
     def foo(self, n):
         return n + 1
-
     @property
-    @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
-    def this_property(self):
-        return 2
+    @deprecated(since=0.9, removal=0.10, alternative="another", is_function=True)
 
 
 Build and write documentation
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -140,15 +140,15 @@ with a possible alternative to be used.
 The decorator should be placed right above the object signature to be deprecated::
 
 .. code-block:: python
-    >>> from diffsims.utils._deprecated import deprecate
-    >>> @deprecate(since=0.8, removal=0.9, alternative="bar")
-    >>> def foo(self, n):
-    >>>     return n + 1
+    from diffsims.utils._deprecated import deprecate
+    @deprecate(since=0.8, removal=0.9, alternative="bar")
+    def foo(self, n):
+        return n + 1
 
-    >>> @property
-    >>> @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
-    >>> def this_property(self):
-    >>>     return 2
+    @property
+    @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
+    def this_property(self):
+        return 2
 
 
 Build and write documentation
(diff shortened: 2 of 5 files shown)


CSSFrancis:
@hakonanes Just adding these here so I can deprecate some code before tackling #198.

Should this be a Gist? or a small python package or??? I really like reusing this :)

hakonanes:
> Should this be a Gist?

Perhaps. Would that make it easier to copy, somehow? Or do you mean in terms of visibility?

As mentioned before, I took the bits I wanted from Matplotlib's and scikit-image's deprecation tools. @harripj improved it after I added it to orix. Did you, @CSSFrancis, add functionality as well at some point? You have my blessing to put it up as a Gist if you want. But, you need @harripj's blessing as well. If you do make a Gist of the these deprecation tools, I'm OK with you licensing it in accordance with Matplotlib and scikit-image, that is not being restricted by GPLv3.

I don't think we should put a package on PyPI, as there are already at least deprecation packages on there:
* deprecation: https://pypi.org/project/deprecation/
* Deprecated: https://pypi.org/project/Deprecated/

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Unreleased
 Added
 -----
 - Explicit support for Python 3.11.
+- Added deprecation tools for deprecating functions and arguments.
```
hakonanes:
```suggestion
- Added deprecation tools for deprecating functions, parameters, methods, and properties.
```

CONTRIBUTING.rst
```diff
@@ -124,6 +124,29 @@ Useful hints on testing:
...
+.. code-block:: python
+    >>> from diffsims.utils._deprecated import deprecate
+    >>> @deprecate(since=0.8, removal=0.9, alternative="bar")
+    >>> def foo(self, n):
+    >>>     return n + 1
+
+    >>> @property
+    >>> @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
+    >>> def this_property(self):
+    >>>     return 2
```
hakonanes: These are code lines in a script rather than lines entered in a prompt, right? I suggest to remove the prompt.

```suggestion
    from diffsims.utils._deprecated import deprecate
    @deprecate(since=0.8, removal=0.9, alternative="bar")
    def foo(self, n):
        return n + 1

    @property
    @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
    def this_property(self):
        return 2
```

diffsims/utils/_deprecated.py
```diff
@@ -0,0 +1,154 @@
...
+# along with diffsims.  If not, see <http://www.gnu.org/licenses/>.
+
+
+"""Helper functions and classes for managing diffsims.
+This module and documentation is only relevant for diffsims developers,
+not for users.
+.. warning:
+    This module and its submodules are for internal use only.  Do not
+    use them in your own code. We may change the API at any time with no
+    warning.
```
hakonanes: Add blank lines for increased readability:

```suggestion
"""Helper functions and classes for managing diffsims.

This module and documentation is only relevant for diffsims developers,
not for users.

.. warning:
    This module and its submodules are for internal use only.  Do not
    use them in your own code. We may change the API at any time with no
    warning.
```

diffsims/utils/_deprecated.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def __init__(
+        self,
+        since: Union[str, int, float],
+        alternative: Optional[str] = None,
+        alternative_is_function: bool = True,
+        removal: Union[str, int, float, None] = None,
+    ):
+        """Visible deprecation warning.
+        Parameters
```
hakonanes:
```suggestion
        """Visible deprecation warning.
        
        Parameters
```

diffsims/utils/_deprecated.py
```diff
@@ -0,0 +1,154 @@
...
+from typing import Callable, Optional, Union
+import warnings
+
+import numpy as np
+
+
+class deprecated:
+    """Decorator to mark deprecated functions with an informative
+    warning.
+    Adapted from
```
hakonanes:
```suggestion
    warning.

    Adapted from
```

diffsims/utils/_deprecated.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        wrapped.__doc__ = new_doc
+
+        return wrapped
+
+
+class deprecated_argument:
+    """Decorator to remove an argument from a function or method's
+    signature.
+    Adapted from `scikit-image
```
hakonanes:
```suggestion
    signature.
    
    Adapted from `scikit-image
```

hakonanes (approved):
Looks good. I have a few minor suggestions that could be addressed before merging.

harripj:
@hakonanes @CSSFrancis you have my blessing too. I wonder whether it is worth adding a reference back to orix somewhere in the Gist (header?)? I think that would be ample accreditation.

hakonanes:
Can you make the suggested changes? I'll have a final look after that. Fairly certain we can merge after that again. I'd like to deprecate some old methods to get sets of {hkl} in #213.

CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
Failing build is due to a Checksum not matching when using pip to install.  Kind of a new one :)

CHANGELOG.rst
hakonanes: It appears you added a new line instead of replacing the current one?

CONTRIBUTING.rst
```diff
@@ -128,6 +128,29 @@ Useful hints on testing:
...
+The decorator should be placed right above the object signature to be deprecated::
+
+.. code-block:: python
+    from diffsims.utils._deprecated import deprecate
+    @deprecate(since=0.8, removal=0.9, alternative="bar")
+    def foo(self, n):
+        return n + 1
+
+    @property
+    @deprecate(since=0.9, removal=0.10, alternative="another", is_function=True)
```
hakonanes:
```suggestion
A deprecation decorator should be placed right above the object signature to be deprecated.

.. code-block:: python
    from diffsims.utils._deprecated import deprecated

    @deprecated(since=0.8, removal=0.9, alternative="bar")
    def foo(self, n):
        return n + 1

    @property
    @deprecated(since=0.9, removal=0.10, alternative="another", is_function=True)
```

hakonanes (changes requested):

CONTRIBUTING.rst
CSSFrancis: I guess I can competently use git some days...

hakonanes (approved):

PR #201, opened 2023-10-25
Use phase class

CSSFrancis:

This changes the simulations to use the `orix.crystal_map.Phase` class and to use the `ReciporicalLatticeVector` for most everything.

It also changes how simulations are set up so that some `Phase` and `Rotation` can be passed to the `SimulationGenerator` object and a list of simulations are returned.  

#### Progress of the PR
- [x] Add `SimulationLibrary`
    - [x] Add Test for `SimulationLibrary`
    - [x] Add plot method for `SimulationLibrary.Rotation`
   ~~- [ ] Add plot method for `SimulationLibrary` (to Hyperspy for visualization?)~~
- [x] Add SimulationLibraries
    - [x] Add good repr
    - [x] Add Tests
- [x] Add SimulationGenerator
- [x] Add Examples
    - [x] Add example for simulating one diffraction pattern
    - [x] Add example for simulating multiple diffraction patterns from one `Phase`
    - [x] Add example for simulating multiple diffraction patterns from multiple `Phase` objects
- [x] Deprecation
    - [x] Deprecate DiffractionGenerator
    - [x] Deprecate LibraryGenerator
    - [ ] Deprecate StrucutureLibrary
    - [x] Deprecate StrucutureLibraryGenerator

```python
####################################
# Simulating One Diffraction Pattern
# ==================================

from diffpy.structure import Atom, Lattice, Structure

from diffsims.generators.simulation_generator import SimulationGenerator

a = 5.431
latt = Lattice(a, a, a, 90, 90, 90)
atom_list = []
for coords in [[0, 0, 0], [0.5, 0, 0.5], [0, 0.5, 0.5], [0.5, 0.5, 0]]:
    x, y, z = coords[0], coords[1], coords[2]
    atom_list.append(Atom(atype="Si", xyz=[x, y, z], lattice=latt))  # Motif part A
    atom_list.append(Atom(
            atype="Si", xyz=[x + 0.25, y + 0.25, z + 0.25], lattice=latt
        )
    )  # Motif part B
struct = Structure(atoms=atom_list, lattice=latt)
p = Phase(structure=struct, space_group=227)

gen = SimulationGenerator(accelerating_voltage=200,)
rot = Rotation.from_axes_angles([1, 0, 0], 45, degrees=True)  # 45 degree rotation around x-axis
sim = gen.calculate_ed_data(phase=p, rotation=rot)

sim.simulations[0].plot()  # plot the first (and only) diffraction pattern

#%%
```

This supercedes #197 
#### For reviewers

      in `.zenodo.json`.


hakonanes:
I'm happy to see that you use `Phase` and `Rotation` directly! These classes should have the necessary visualization tools required for validating the phase and rotation list. Please let me know if you have any questions in this regard. I'm more than happy to help with changes to these classes that make sense in the context of orix.

One issue is related to the rotation sampling methods, which I made a comment about in the PR which this PR supersedes: https://github.com/pyxem/diffsims/pull/197#issuecomment-1779503727.

CSSFrancis:
> I'm happy to see that you use `Phase` and `Rotation` directly! These classes should have the necessary visualization tools required for validating the phase and rotation list. Please let me know if you have any questions in this regard. I'm more than happy to help with changes to these classes that make sense in the context of orix.

@hakonanes how does the `Rotation` visualization work?  I've been trying to figure that out but can't seem to visualize the different rotations.  Can I just force them to be Vector3D object somehow and then plot the stereographic projection from there?

CSSFrancis:
Okay I've decided a couple of things are kind of anti patterns in `diffsims` and I don't really like the shear number of classes/ generators etc. 

I've come to a "One class to rule them all" mentality similar to the `BaseSignal` class in hyperspy.  There are really 3 different situations we want to handle.

1. One `Phase` and one `Rotation`
2. One `Phase` and multiple `Rotation`(s)
3.  Multiple `Phase`(s) and multiple `Rotation`(s)

All of them are useful and we would like to be able to template match vs all of these options for

 1--> Orientation mapping on some zone axis with a tilt boundary 
2 --> Orientation mapping on a crystal with multiple orientations but a single phase 
3 -->   Orientation mapping on a crystal with multiple orientations and multiple phases

That being said the output should be the same (a `CrystalMap` object).  Further there really isn't need for more than one generator object (a `DiffractionGenerator`) 

So that leads us to a `Simulation` object which has 3 levels of organization.

`phases` --> `rotations` -->`coordinates`  where each phase has some rotations and each rotation has some coordinates. 

Let's see what that means:

```python

sim # simulation
sim.iphase["al"] # Returns only the simulation of the Al phase
sim.irot[0:4] # Returns only the first 4 rotations for all phases

sim.plot() # plots the simulation at sim.phase_index, sim.rot_index
``` 

So you can do things like...
```python
sim.iphase["al"].irot[2].plot()
```

It also gives us the potential to add in things like interactive plotting if we want to without much extra work. The fact that everything is the same object gives us lots of flexibility when it comes to creating simulations, slicing them etc.  It also makes it possible for them to grow a little bit and we could include some more complexity/ extend them as well. 

@hakonanes any thoughts? I have a working class in this PR under `diffsims.simulations.Simulation`

CSSFrancis:
@hakonanes What exactly is the check-Manifest test (and why is it failing :))

hakonanes:
It matches all files listed in MANIFEST.in intended for inclusion in the source distribution to an actual distribution it creates. It fails if there is a mismatch.

The MANIFEST.in file ensures that the source dist contains just what we intend and nothing else.

CSSFrancis:
Just for reference we should be able to use something like https://github.com/hyperspy/hyperspy/issues/3271 for visualization.  I think that would be better than trying to reimplement the visualization tools here.

pc494:
I'm happy to take on the review of this.

CSSFrancis:
@pc494 that would be great! I haven't changed the 1 d simulation but that can be done in a different pr

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ Unreleased
 Added
 -----
 - Explicit support for Python 3.11.
+- Added deprecation tools for deprecating functions and arguments.
 
 Changed
 -------
```
pc494, line 26: This will need to be filled out.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 23.1.0
+    hooks:
+      - id: black
+      - id: black-jupyter
+        files: \.ipynb
+        args: [--line-length=77]
```
pc494: Happy(ish) with this being the default in all of the family of packages. Can I just confirm that it doesn't add much in the way of a barrier to entry for new users?

CONTRIBUTING.rst
```diff
@@ -124,6 +124,29 @@ Useful hints on testing:
...
   <https://doc.pytest.org/en/latest/how-to/parametrize.html>`_.
 
+
+Deprecations
+------------
+We attempt to adhere to semantic versioning as best we can. This means that as little,
+ideally no, functionality should break between minor releases. Deprecation warnings
+are raised whenever possible and feasible for functions/methods/properties/arguments,
+so that users get a heads-up one (minor) release before something is removed or changes,
+with a possible alternative to be used.
```
pc494: Would make this a bit shorter. I was thinking:

'This project attempts to adhere to (pre 1.0.0) semantic versioning and ideally no functionality should be broken between minor releases. Deprecation warnings are raised to give user a full minor release cycle to adjust their code before a breaking change is implemented.'

diffsims/simulations/simulation.py
```diff
@@ -0,0 +1,534 @@
...
+from orix.crystal_map import Phase
+from orix.quaternion import Rotation
+from orix.vector import Vector3d
+
+from diffsims.crystallography.reciprocal_lattice_vector import ReciprocalLatticeVector
+from diffsims.pattern.detector_functions import add_shot_and_point_spread
+from diffsims.utils import mask_utils
+
+if TYPE_CHECKING:  # pragma: no cover
+    from diffsims.generators.simulation_generator import SimulationGenerator
```
pc494: Could you explain the idea behind this?

diffsims/simulations/simulation.py
```diff
@@ -0,0 +1,534 @@
+from typing import Union, Sequence, TYPE_CHECKING, Any
```
pc494: Copyright statement will be needed here.

diffsims/simulations/__init__.py
```diff
@@ -0,0 +1 @@
+from diffsims.simulations.simulation import Simulation
```
pc494: Copyright statement should probably be here.

diffsims/tests/simulations/test_simulation.py
```diff
@@ -0,0 +1,350 @@
+import numpy as np
```
pc494, line 19: might be worth running a find replace to catch all the missing copyrights?

diffsims/utils/_deprecated.py
```diff
@@ -0,0 +1,154 @@
...
+    Adapted from
+    `scikit-image
+    <https://github.com/scikit-image/scikit-image/blob/main/skimage/_shared/utils.py>`_
+    and `matplotlib
+    <https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/_api/deprecation.py>`_.
+    """
+
+    def __init__(
+        self,
+        since: Union[str, int, float],
```
pc494, line 49: this feels to generous as a typing. Surely string is enough?

examples/creating_a_simulation_library/simulating_diffraction_patterns.py
```diff
@@ -0,0 +1,91 @@
...
+
+sim.irot[3].plot(show_labels=True)  # plot the fourth(45 degrees) diffraction pattern
+# %%
+
+sim.coordinates  # coordinates of all the diffraction patterns
+
+# ============================================================
+# Simulating Multiple Rotations for Multiple Phases
+# ============================================================
+p2 = p.deepcopy()
```
pc494: might be worth a comment about the use of .deepcopy() here?

pc494 (commented):
This is generally a great PR making a much-needed change. It does still need a bit of cleaning up though.

If @hakonanes has the capacity to look into some of the more technical bits of the implementation, I didn't spend that much time working directly with `Phase`.

diffsims/simulations/simulation.py
CSSFrancis: The problem is that you would have a circular import if you didn't have that. It's a work around to allow the type checking to work. I'm still kind of hesitant with type checking because I do think it makes things a little more complicated for new users. That being said it really helps IDEs and ai generative text so...

I'll add a comment explaining why

.pre-commit-config.yaml
CSSFrancis: Yea that's actually the beauty of the pre commit CI. 1 you can set up pre-commit on your own computer which works the best or 2 we can just run the precommit by commenting "precommit autofix". 

Black is one of those things you either have to go all in on or not. If half of your developers are using it you help up with terribly confusing git histories.

CSSFrancis: But I have the same sentiments. Like everyone I kind of dislike black but dislike the alternative more.

pc494: I think any/all steps to avoid massive 'run black on everything' commits is an improvement. Marking as resolved.

diffsims/simulations/simulation.py
pc494: Happy with this.

CSSFrancis:
@hakonanes any chance you can look over the Phase implementation here?

hakonanes:
I'll review it this weekend. Thank you for pinging me. And your patience (who'd have thought that a new job, relocating, and PhD defence coming up would mean such busy times...).

CSSFrancis:
@hakonanes No problem, thanks for reviewing this! 

I'm probably 3 months behind you, so I'm going to be there very very soon.  Just trying to finish some stuff up before I'm done.

pc494:
sorry @hakonanes, this one is on me...

hakonanes:
As in you'll review it? Perfect!

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,370 @@
...
+    def __init__(
+        self,
+        accelerating_voltage: float = 200,
+        scattering_params: str = "lobato",
+        precession_angle: float = 0,
+        shape_factor_model: str = "lorentzian",
+        approximate_precession: bool = True,
+        minimum_intensity: float = 1e-20,
+        **kwargs,
+    ):
```
pc494, line 76: This should be documented. Although some units will be clear (`accelerating_voltage`) others will probably not be.

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,370 @@
...
+        max_excitation_error
+            The cut-off for geometric excitation error in the z-direction
+            in units of reciprocal Angstroms. Spots with a larger distance
+            from the Ewald sphere are removed from the pattern.
+            Related to the extinction distance and roungly equal to 1/thickness.
+        shape_factor_width
+            Determines the width of the reciprocal rel-rod, for fine-grained
+            control. If not set will be set equal to max_excitation_error.
+        debye_waller_factors
+            Maps element names to their temperature-dependent Debye-Waller factors.
```
pc494: This also should have types as the typing in the args is afaik optional.

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,370 @@
...
+        ----------
+        phase:
+            The phase for which to derive the diffraction pattern.
+        reciprocal_radius
+            The maximum radius of the sphere of reciprocal space to
+            sample, in reciprocal Angstroms.
+        minimum_intensity
+            The minimum intensity of a reflection to be included in the profile.
+        debye_waller_factors
+            Maps element names to their temperature-dependent Debye-Waller factors.
```
pc494, line 260: see above comments about docstrings.

pc494 (changes requested):
This is more or less good to go. @CSSFrancis while you were building this, were there consistency checks that the results agreed with what we had been getting previously? If so might be worth documenting them in the PR. That would leave

- [ ]  Coax the final bits of documentation into agreement with standards.
- [ ] Provide some evidence of sensible simulations (doesn't have to be loads).
 
Then I reckon it's merge and make a new minor release time? :)

CSSFrancis:
> This is more or less good to go. @CSSFrancis while you were building this, were there consistency checks that the results agreed with what we had been getting previously? If so might be worth documenting them in the PR. That would leave

Yea let me add those in.  I realized last night that there are some (small) inconstancies which arose because we aren't calculating the zero beam I think the Vector3D class in orix doesn't love the idea of a [0,0,0] vector. The result is that the thresholding doesn't work properly.  This is kind of like the old method `with_zero_beam=False`

I'll make a migration guide. 


> * [ ]   Coax the final bits of documentation into agreement with standards.

Good point!  I'll crawl through the documentation to make sure that is there

> * [ ]  Provide some evidence of sensible simulations (doesn't have to be loads).

Will do!

> Then I reckon it's merge and make a new minor release time? :)

Yea! Let's see what I can do this weekend to clean up all of this stuff!

hakonanes:
BTW, orix forces a [crystal axes alignment](https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html) that is different from diffpy.structure. Simulations before and after this PR of a crystal with non-right angles, such as a triclinic or monoclinic crystal, should be compared.

CSSFrancis:
@hakonanes hmm I've noticed that the new simulations are mirrored about the x axis. 

Is this a result of this as well?  I'll go ahead and make a comparison for a set of different crystal structures (and we can probably add this as an example :))

CSSFrancis:
@hakonanes Maybe you can answer this question (as I'm not really good at rotations and conventions)

Currently if we do something like:

```python
from transforms3d.euler import euler2mat
from orix.quaternion.rotation import Rotation

rot = Rotation.from_euler([0,45,0], degrees=True)
rot2 = euler2mat(0,45,0, axes="")
rot_vect = lat_v.rotate_from_matrix(rotation_matrix=rot.to_matrix()[0])      
```
The rotations are:

```
[[[ 1.          0.          0.        ]
  [ 0.          0.70710678  0.70710678]
  [ 0.         -0.70710678  0.70710678]]]

[[ 1.          0.         -0.        ]
 [-0.          0.52532199 -0.85090352]
 [ 0.          0.85090352  0.52532199]]
```

I think this is just a matter of convention?

CSSFrancis:
~~So things seem to be equivalent, however, we need to have a better way of defining an equivalent Rotation.  @hakonanes Is this something that we should be able to add to orix/ is there already a way to do this in orix?~~

I deleted a couple of comments that were no longer relevant.  Sorry, so many different ways to define rotations and I kept messing up small things :) 

Here's a comparison of the two methods side by side. 

```python
p = Phase.from_cif("../data/Fe.cif")

sim_gen = SimulationGenerator(200)
dif_gen = DiffractionGenerator(200)


rot = Rotation.from_euler([[0,0,0],[45,45,0]], degrees=True, direction="crystal2lab" )

ed = sim_gen.calculate_ed_data(phase=p, rotation=rot, reciprocal_radius=1)
ed2 = [dif_gen.calculate_ed_data(structure=p.structure, rotation=[0,0,0], reciprocal_radius=1, with_direct_beam=True),
       dif_gen.calculate_ed_data(structure=p.structure, rotation=[45,45,0],reciprocal_radius=1, with_direct_beam=True)]


import matplotlib.pyplot as plt
fig, axs = plt.subplots(2,2)
for i in range(2):
    ed.irot[i].plot(ax=axs[i,0])
    axs[i,0].set_xlim(-1,1)
    axs[i,0].set_ylim(-1,1)
    ed2[i].plot(ax=axs[i,1])
    axs[i,1].set_xlim(-1,1)
    axs[i,1].set_ylim(-1,1)
    
axs[0,0].set_title("New")
axs[0,1].set_title("Old")
```

Figure 1

CSSFrancis:
@hakonanes Also there aren't really any triclinic metals and we don't really support ionic states (We should add this it wouldn't be hard) so I haven't gotten around to trying this...

Edit: I can add this in a different PR

pc494:
> BTW, orix forces a [crystal axes alignment](https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html) that is different from diffpy.structure. Simulations before and after this PR of a crystal with non-right angles, such as a triclinic or monoclinic crystal, should be compared.

@CSSFrancis as a temporary fix, I would accept throwing a warning in this PR and then fixing the behaviour in a later PR (as long as we don't release with the warning still in). If nothing else, this would keep the size of the reviews and PRs down a bit. @hakonanes would that be an acceptable compromise?

rbjorge:
Sticking with a cubic crystal for now, this is what I get with diffsims 0.5.2:
```python
import numpy as np
import diffpy
from diffsims.libraries.structure_library import StructureLibrary
from diffsims.generators.library_generator import DiffractionLibraryGenerator

latt = diffpy.structure.lattice.Lattice(4,4,4,90,90,90)
atoms = [diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.0,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.5,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.0,0.5],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.5,0.5],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
euler_angles = np.array([[10.,0.,0.]])
struct_library = StructureLibrary(["Al"], [structure_matrix], [euler_angles])
diff_gen = DiffractionGenerator(accelerating_voltage=200)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(struct_library,
                                           calibration=0.0262,
                                           reciprocal_radius=1.6768,
                                           half_shape=64,
                                           with_direct_beam=True,
                                           max_excitation_error=0.02)
diff_lib["Al"]["simulations"][0].plot(size_factor=30, show_labels=True);
```
diffsims052_Al_10_0_0

And this is what I get with this PR:
```python
import diffpy
from diffsims.generators.simulation_generator import SimulationGenerator

latt = diffpy.structure.lattice.Lattice(4,4,4,90,90,90)
atoms = [diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.0,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.5,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.0,0.5],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.5,0.5],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
p = Phase("Al", point_group="m-3m", structure=structure_matrix)
gen = SimulationGenerator(accelerating_voltage=200)
rot = Rotation.from_euler([10.,0.,0.], degrees=True)
sim = gen.calculate_ed_data(phase=p, rotation=rot, reciprocal_radius=1.6768, max_excitation_error=0.02, with_direct_beam=True)
sim.plot(size_factor=30, show_labels=True);
```
diffsims_new_Al_10_0_0

As you can see, the interpretation of the Euler angles seems to have changed. Also, labels of the reflections are wrong: the 100 and 300 reflections are kinematically forbidden, while the 200 reflections should be allowed.

CSSFrancis:
@rbjorge Thanks for checking this!  That is a good catch... I'm not sure why the Euler angles interpretation has changed for your example but not when building from the CIF file... @hakonanes might have some more information on why that might be the case.

As for the wrong labels, that is a weird one (that I should have checked!) so thank you for catching that.  I suspect that the `ReciprocalLatticeVector` class is doing something weird when we ask for the hkl values.  I'll have to look into that more.

rbjorge:
It's the same behavior whether `Phase` comes from a CIF file or from `diffpy.structure.Structure`.

I notice that you used `Rotation.from_euler(direction="crystal2lab")` in your example above. Since `DiffractionLibraryGenerator` and `DiffractionGenerator` use the "lab2crystal" convention (like the default in `orix`), things might somehow have gotten inverted twice in your `SimulationGenerator.calculate_ed_data()` example above, making the old and new way look identical?

The `show_labels` option is really useful!

CSSFrancis:
@rbjorge that's likely! I'm by no means a crystallographer. My research is all on amorphous samples :) The lab2crystal and crystal2lab convention confuses me a little bit. Is this assuming the beam direction is in the -z direction? That could be a source of the error.

rbjorge:
Yes, it seems like diffsims currently assumes the beam travels along the `-z` direction. I tried to document these things in [https://pyxem.readthedocs.io/en/stable/tutorials/pyxem-demos/13%20Conventions.html](https://pyxem.readthedocs.io/en/stable/tutorials/pyxem-demos/13%20Conventions.html), to make things less confusing. In the plots above, the x- and y-axes suggest that the z-axis is pointing out of the screen. Is the beam traveling out of the screen?

We are beginning to use absolute orientations here in Trondheim (for determining the holder tilt axes), and it's therefore nice to have somewhat stable and defined conventions.

What adds to the confusion is maybe that the functions that generate different crystal orientations (`diffsims.generators.rotation_list_generators.get_beam_directions_grid()` and `diffsims.generators.sphere_mesh_generators.beam_directions_grid_to_euler()`)
use the term "beam directions", even though the beam is fixed with respect to the camera. To me, different beam directions would mean different beam tilts (with respect to the optical axis).

By the way, I also see that you extended the `ReciprocalLatticeVector` class with a `rotate_from_matrix()` function. I don't think I quite understand how it works. It might be worth looking into when determining the source of the discrepancy.

CSSFrancis:
@rbjorge It seems the problem is that we weren't appropriately accounting for the coordinate --> hkl transform with the rotation.  We should probably be rotating the entire lattice as well as the reciprocal space coordinates.  There isn't a way to do this easily in diffsims (It would be a good thing to add!)

I just rotated things the hkl values as well and now things align much better. 

As far as the mirroring.  I might need some help :) I can just flip the values to match but I don't know if I completely understand the scope of the problem.

rbjorge:
I notice that `DiffractionGenerator.calculate_ed_data()` uses `transforms3d.euler`. This package seems to interpret Euler angles such that the rotation matrix is transposed (or you could say inverted since it's an orthogonal matrix) compared with `orix`:

```python
import numpy as np
from transforms3d.euler import euler2mat

a=23.; b=24.; c=25.
mat = euler2mat(np.deg2rad(a), np.deg2rad(b), np.deg2rad(c), axes="rzxz")
rot = Rotation.from_euler([a,b,c], degrees=True)
mat_orix = rot.to_matrix()[0]
print(mat)
print(mat_orix)
print(np.allclose(mat, mat_orix.T))

[[ 0.68340685 -0.71252931  0.15892467]
 [ 0.70951209  0.59700498 -0.37440305]
 [ 0.17189433  0.36862859  0.91354546]]

[[ 0.68340685  0.70951209  0.17189433]
 [-0.71252931  0.59700498  0.36862859]
 [ 0.15892467 -0.37440305  0.91354546]]

True
```

Now checking with `MTEX`, which confusingly uses "lab2crystal" for Euler angles and "crystal2lab" for rotation matrices, I find:
```matlab
rotation.byEuler(23.*degree, 24.*degree, 25.*degree).matrix

ans =

    0.6834   -0.7125    0.1589
    0.7095    0.5970   -0.3744
    0.1719    0.3686    0.9135
```
So I would say that `orix` interprets Euler angles the way `MTEX` does.

So, without having dug deep into the 2 different `calculate_ed_data()` functions, I would guess that the `DiffractionGenerator` one gets the rotation matrix in a "crystal2lab" "format", and then somehow inverts this to give the same result as ASTAR Diffgen (which uses a "lab2crystal" convention, as does [ReciPro](https://seto77.github.io/ReciPro/) which is free). So maybe, when switching to using `orix` for rotations in `SimulationGenerator` you get a "lab2crystal" plot when using `direction="crystal2lab"`, for similar reasons?

CSSFrancis:
> So, without having dug deep into the 2 different `calculate_ed_data()` functions, I would guess that the `DiffractionGenerator` one gets the rotation matrix in a "crystal2lab" "format", and then somehow inverts this to give the same result as ASTAR Diffgen (which uses a "lab2crystal" convention, as does [ReciPro](https://seto77.github.io/ReciPro/) which is free). So maybe, when switching to using `orix` for rotations in `SimulationGenerator` you get a "lab2crystal" plot when using `direction="crystal2lab"`, for similar reasons?

@rbjorge That sounds completely reasonable and is almost certainly the case.  Now the question is do we keep things as they are currently or follow `orix`?  Is there anything wrong with the new implementation or is it just different? 

I obviously don't want to break workflows etc. but not following `orix` seems like we will be fighting a losing battle and I don't really want to do that.

rbjorge:
I would say keep the current behavior of diffsims *and* use orix.

Looking a bit at the current behavior of `DiffractionGenerator.calculate_ed_data()` and in particular [where the reflections are rotated](https://github.com/pyxem/diffsims/blob/fae5ca573b6ecb521d0f11fdce0f6a1e1695f6b6/diffsims/generators/diffraction_generator.py#L254). I think the issue is that we are really rotating a vector in the crystal reference frame, so when using "lab2crystal" the rotation needs to be inverted. So, for example, looking at the (200) reflection in an fcc lattice (and using `orix` instead of `euler2mat`):

```python
import numpy as np
import diffpy
from diffsims.utils.sim_utils import get_points_in_sphere
latt = diffpy.structure.lattice.Lattice(4,4,4,90,90,90)
recip_latt = latt.reciprocal()
g_indices, cartesian_coordinates, g_hkls = get_points_in_sphere(recip_latt, reciprocal_radius=1.6768)
print(g_indices[888])
print(g_hkls[888])
print(cartesian_coordinates[888])
rot = Rotation.from_euler([10., 0., 0.], degrees=True).to_matrix()[0]
print(np.matmul(rot, cartesian_coordinates[888]))
print(np.matmul(rot.T, cartesian_coordinates[888]))

[2. 0. 0.]
0.5
[0.5 0.  0. ]
[ 0.49240388 -0.08682409  0.        ]
[0.49240388 0.08682409 0.        ]
```

Without the transpose, the (200) reflection would be in the 4th quadrant. With the transpose it is in the 1st quadrant, where I think it should be.

If this is the source of the discrepancy, then the rotations in your `SimulationGenerator` should be transposed to get the same result as `DiffractionGenerator`. I guess ideally it should be possible to do `~Rotation * ReciprocalLatticeVector` instead of using `rotate_from_matrix()`?

CSSFrancis:
```python
"""
Migration Guide
===============
This is a migration guide for version 0.5.x to 0.6.x. This guide helps to show the changes
that were made to the API and how to update your code to use the new API.
"""

# Old

import numpy as np
import diffpy
from diffsims.libraries.structure_library import StructureLibrary
from diffsims.generators.library_generator import DiffractionLibraryGenerator


latt = diffpy.structure.lattice.Lattice(4,4,4,90,90,90)
atoms = [diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.0,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.5,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.0,0.5],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.5,0.5],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
euler_angles = np.array([[0,0,0],[10.,0.,0.]])
struct_library = StructureLibrary(["Al"], [structure_matrix], [euler_angles])
diff_gen = DiffractionGenerator(accelerating_voltage=200)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(struct_library,
                                           calibration=0.0262,
                                           reciprocal_radius=1.6768,
                                           half_shape=64,
                                           with_direct_beam=True,
                                           max_excitation_error=0.02)


# New


import diffpy
from diffsims.generators.simulation_generator import SimulationGenerator

latt = diffpy.structure.lattice.Lattice(4,4,4,90,90,90)
atoms = [diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.0,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.5,0.0],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.5,0.0,0.5],lattice=latt),
         diffpy.structure.atom.Atom(atype='Al',xyz=[0.0,0.5,0.5],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
p = Phase("Al", point_group="m-3m", structure=structure_matrix)
gen = SimulationGenerator(accelerating_voltage=200)
rot = Rotation.from_euler([[0,0,0],[10.,0.,0.]], degrees=True)
sim = gen.calculate_ed_data(phase=p, rotation=rot, reciprocal_radius=1.6768, max_excitation_error=0.02, with_direct_beam=True)
sim.plot(size_factor=30, show_labels=True);


import matplotlib.pyplot as plt

fig, axs = plt.subplots(2, 2)
for i in range(2):
    diff_lib["Al"]["simulations"][i].plot(size_factor=15, show_labels=True, ax=axs[i, 0]);
    sim.irot[i].plot(ax=axs[i, 1],size_factor=15, show_labels=True, la)
    axs[i, 0].set_xlim(-1.5, 1.5)
    axs[i, 0].set_ylim(-1.5, 1.5)
    axs[i, 1].set_xlim(-1.5, 1.5)
    axs[i, 1].set_ylim(-1.5, 1.5)

axs[0, 0].set_title("New")
axs[0, 1].set_title("Old")
```

image

CSSFrancis:
@pc494 What do you think about trying to get this merged?  I think we should clean up the documentation a bit and there might be some small changes to get it to work well with the orientation mapping but I do think we've got a good start here.

rbjorge:
> If this is the source of the discrepancy, then the rotations in your `SimulationGenerator` should be transposed to get the same result as `DiffractionGenerator`. I guess ideally it should be possible to do `~Rotation * ReciprocalLatticeVector` instead of using `rotate_from_matrix()`?

By the way, I see that this is also how `orix`'s [`_plot_unit_cell()`](https://github.com/pyxem/orix/blob/main/orix/plot/unit_cell_plot.py#L71) works.

rbjorge:
For non-orthogonal crystal structures, the nice thing about using the `Phase` class is that it would resolve pyxem/pyxem#947. But checking with a hexagonal structure in the $[100]$ zone axis I get with diffsims 0.5.2 (where $X||a^*$):
```python
import numpy as np
import diffpy
from diffsims.libraries.structure_library import StructureLibrary
from diffsims.generators.library_generator import DiffractionLibraryGenerator

latt = diffpy.structure.lattice.Lattice(2.464,2.464,6.711,90,90,120)
atoms = [diffpy.structure.atom.Atom(atype='C',xyz=[0.0,0.0,0.25],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.0,0.0,0.75],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.333333,0.666667,0.25],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.666667,0.333333,0.75],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
euler_angles = np.array([[0.,90.,120.]])
struct_library = StructureLibrary(["Graphite"], [structure_matrix], [euler_angles])
diff_gen = DiffractionGenerator(accelerating_voltage=200)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(struct_library,
                                           calibration=0.0262,
                                           reciprocal_radius=1.6768,
                                           half_shape=64,
                                           with_direct_beam=True,
                                           max_excitation_error=0.02)
diff_lib["Graphite"]["simulations"][0].plot(size_factor=30, show_labels=True);
```
diffsims_v0_5_2_graphite_0_90_120
Because of the different choice of orthogonal unit cell ($X||a$), the $[100]$ zone axis is (0&deg;, 90&deg;, 90&deg;) in orix. With this PR I get:
```python
import diffpy
from diffsims.generators.simulation_generator import SimulationGenerator

p = Phase("Graphite", point_group="6/mmm", structure=structure_matrix)  # defined above
gen = SimulationGenerator(accelerating_voltage=200)
rot = Rotation.from_euler([0.,90.,90.], degrees=True)
sim = gen.calculate_ed_data(phase=p, rotation=rot, reciprocal_radius=1.6768, max_excitation_error=0.02, with_direct_beam=True)
sim.plot(size_factor=30, show_labels=True);
```
diffsims_PR#201_graphite_0_90_90

Something seems to be off, both with the labels, and the difference between the $00l$ row and the other rows. Without rotation, the patterns look as expected, so I guess there is something about the rotation of the reciprocal lattice vectors.

Here is also the output from ASTAR Diffgen, which uses the same convention as orix. It looks the same as $[100]$ in diffsims 0.5.2.
graphite_0_90_90

CSSFrancis:
@rbjorge, Ugh your totally right!   

This is frustrating and I feel like I need to be much better at this.  I guess part of the problem is that I am not sure __how__ the `ReciporicalLatticeVector` class is defined.  I assumed that the `data` attribute is the cartesian coordinates in reciprocal space.  Then to get something like the hkl you would have to  rotate back, take the known lattice affine transformation, and then matrix multiply by the transpose of the  lattice affine transformation.  My guess is that we are doing it out of order right now which is what is messing things up. 


Actually if we do that we get the answer below.  This at least seems to solve the labels problem but for some reason we are missing some of the weaker diffraction spots.

Screenshot 2024-02-05 at 8 57 08 AM

CSSFrancis:
There doesn't seem to be a reason for them to be missing as the methods should be almost identical.

Maybe the `from_min_d_spacing` method isn't picking them up?  https://github.com/pyxem/diffsims/blob/fae5ca573b6ecb521d0f11fdce0f6a1e1695f6b6/diffsims/crystallography/reciprocal_lattice_vector.py#L1073C9-L1131

CSSFrancis:
@rbjorge What do you think?  Let me know if you can find any more differences :). There __shouldn't__ be...

```python
import diffpy
import matplotlib.pyplot as plt
from diffsims.libraries.structure_library import StructureLibrary
from diffsims.generators.library_generator import DiffractionLibraryGenerator

latt = diffpy.structure.lattice.Lattice(2.464,2.464,6.711,90,90,120)
atoms = [diffpy.structure.atom.Atom(atype='C',xyz=[0.0,0.0,0.25],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.0,0.0,0.75],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.333333,0.666667,0.25],lattice=latt),
         diffpy.structure.atom.Atom(atype='C',xyz=[0.666667,0.333333,0.75],lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms,lattice=latt)
euler_angles = np.array([[0.,90.,120.]])
struct_library = StructureLibrary(["Graphite"], [structure_matrix], [euler_angles])
diff_gen = DiffractionGenerator(accelerating_voltage=200)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(struct_library,
                                           calibration=0.0262,
                                           reciprocal_radius=1.6768,
                                           half_shape=64,
                                           with_direct_beam=True,
                                           max_excitation_error=0.02)

import diffpy
from diffsims.generators.simulation_generator import SimulationGenerator

p = Phase("Graphite", point_group="6/mmm", structure=structure_matrix)  # defined above
gen = SimulationGenerator(accelerating_voltage=200)
rot = Rotation.from_euler([0.,90.,90.], degrees=True)
sim = gen.calculate_ed_data(phase=p, rotation=rot, reciprocal_radius=1.6768, max_excitation_error=0.02, with_direct_beam=True,)


fig, axs = plt.subplots(1,2, figsize=(12,6))


sim.plot(size_factor=30, show_labels=True, min_label_intensity=0.09,ax=axs[1]);
diff_lib["Graphite"]["simulations"][0].plot(size_factor=30, show_labels=True, ax=axs[0]);

axs[0].set_ylim(-1.5, 1.5)
axs[0].set_xlim(-1.5, 1.5)

axs[1].set_ylim(-1.5, 1.5)
axs[1].set_xlim(-1.5, 1.5)
```
image

rbjorge:
> @rbjorge What do you think? Let me know if you can find any more differences :). There **shouldn't** be...
> 
It looks good to me.

rbjorge:
I often use different `max_excitation_error`s for different phases in `calculate_ed_data()` (when template matching multiple phases in  a dataset, e.g., particles in a matrix). Currently I can run  `diff_lib.update(new_phase)` since `diffsims.libraries.diffraction_library.DiffractionLibrary` is a subclass of `dict`. Is something similar possible with the [`diffsims.simulations.simulation2d.Simulation2D`](https://github.com/CSSFrancis/diffsims/blob/use_phase_class/diffsims/simulations/simulation2d.py#L102) class that you use in this PR (or do you plan to make a `SimulationLibrary` class)?

CSSFrancis:
> I often use different `max_excitation_error`s for different phases in `calculate_ed_data()` (when template matching multiple phases in a dataset, e.g., particles in a matrix). Currently I can run `diff_lib.update(new_phase)` since `diffsims.libraries.diffraction_library.DiffractionLibrary` is a subclass of `dict`. Is something similar possible with the [`diffsims.simulations.simulation2d.Simulation2D`](https://github.com/CSSFrancis/diffsims/blob/use_phase_class/diffsims/simulations/simulation2d.py#L102) class that you use in this PR (or do you plan to make a `SimulationLibrary` class)?

@rbjorge It isn't currently possible. My idea was to just have one class for everything with the idea that when you want to do something like template matching you usually just want a long list of simulation objects.  It's definitely something that we can add if it's wanted.  We can also just have it so if the max_excitation_error is  passed as a list then different values are used for each.  I think that is an easier place to start!

viljarjf:
>  Let me know if you can find any more differences 

Other than {0 0 X} (for all X), the intensities still differ, e.g. {0 1 1} is much larger in the right plot than the left.

The coordinate definitions are also still different, but this might be intended? 
At least, using euler angles (0, 90, 90) for both I get the following:
bilde

CSSFrancis:
@rbjorge @viljarjf I think this difference is related to the change from diffsims --> orix correct?  If you plot (90,90,0) and (90, 120,0) The values should be almost the same?

@rbjorge do you have a better understanding of the difference?

viljarjf:
With [0, 90, 120] in the old simulation, and [0, 90, 90] in the new, as you had in your final example, the positions are the same, but the intensities are still different. I have not looked into why

CSSFrancis:
The intensity difference is fairly minor and a result of changing how the indices are calculated or related to the rotation being applied (possibly a floating point error?)  Honestly does it matter? I checked and the difference is at maximum `6.665895211753391e-11` inverse angstroms difference. 

We are going to have to break somethings to get this working... 

Just to give you an idea.  The changes are something like:

`[-9.37256931e-01, -7.45045448e-01,  2.91816436e-16] -->[-9.37256931e-01, -7.45045448e-01,  3.73546167e-16]`

Not sure how to handle this...

viljarjf:
I get a fairly large difference in intensities for the reflections with two non-zero indices. E.g. for [0 1 2], the difference is 0.0044 -> 0.0495, i.e over 10x increase:
~~~python
import diffpy
import numpy as np
import matplotlib.pyplot as plt
from diffsims.libraries.structure_library import StructureLibrary
from diffsims.generators.library_generator import DiffractionLibraryGenerator
from diffsims.generators.simulation_generator import SimulationGenerator

# Shared parameters
latt = diffpy.structure.lattice.Lattice(2.464, 2.464, 6.711, 90, 90, 120)
atoms = [diffpy.structure.atom.Atom(atype='C', xyz=[0.0, 0.0, 0.25], lattice=latt),
         diffpy.structure.atom.Atom(atype='C', xyz=[0.0, 0.0, 0.75], lattice=latt),
         diffpy.structure.atom.Atom(atype='C', xyz=[1/3, 2/3, 0.25], lattice=latt),
         diffpy.structure.atom.Atom(atype='C', xyz=[2/3, 1/3, 0.75], lattice=latt)]
structure_matrix = diffpy.structure.Structure(atoms=atoms, lattice=latt)
euler_angles_old = np.array([[0, 90, 120]])
euler_angles_new = np.array([[0, 90, 90]])
calibration = 0.0262
reciprocal_radius = 1.6768
with_direct_beam = False
max_excitation_error = 0.02
accelerating_voltage = 200
shape = (64, 64)
sigma = 1.4

# Old
struct_library = StructureLibrary(["Graphite"], [structure_matrix], [euler_angles_old])
diff_gen = DiffractionGenerator(accelerating_voltage=accelerating_voltage)
lib_gen = DiffractionLibraryGenerator(diff_gen)
diff_lib = lib_gen.get_diffraction_library(struct_library,
                                           calibration=calibration,
                                           reciprocal_radius=reciprocal_radius,
                                           with_direct_beam=with_direct_beam,
                                           max_excitation_error=max_excitation_error,
                                           half_shape=shape[0]//2,
                                           )

# New
p = Phase("Graphite", point_group="6/mmm", structure=structure_matrix)
gen = SimulationGenerator(accelerating_voltage=accelerating_voltage)
rot = Rotation.from_euler(euler_angles_new, degrees=True)
sim = gen.calculate_ed_data(phase=p, 
                            rotation=rot, 
                            reciprocal_radius=reciprocal_radius, 
                            max_excitation_error=max_excitation_error, 
                            with_direct_beam=with_direct_beam,
                            )

# Compare
old_data = diff_lib["Graphite"]["simulations"][0].get_diffraction_pattern(shape=shape, sigma=sigma)
new_data = sim.get_diffraction_pattern(shape=shape, sigma=sigma, calibration=calibration)

plt.figure(figsize=(9.5, 3))
plt.subplot(1, 3, 1)
plt.title("Difference |new - old|")
plt.imshow(np.abs(new_data - old_data), vmax=0.05)
plt.axis("off")
plt.colorbar()
plt.subplot(1, 3, 2)
plt.title("Old simulation, vmax=0.05")
plt.imshow(old_data, vmax=0.05)
plt.axis("off")
plt.colorbar()
plt.subplot(1, 3, 3)
plt.title("New simulation, vmax=0.05")
plt.imshow(new_data, vmax=0.05)
plt.axis("off")
plt.colorbar()
plt.tight_layout()
~~~
bilde

CSSFrancis:
@viljarjf Huh all from a change of:

 `[-4.68628465e-01, -2.98018179e-01,  1.70229751e-16] -->[-4.68628465e-01, -2.98018179e-01,  1.41346131e-16]`

I guess we have to think about if this actually matters.  There isn't anything wrong with either simulation I think, but it's just something that we might have to handle.   For template matching I actually don't think it will actually make a huge difference....

But there isn't a way to fix this discrepancy really.  The problem is probably somewhere buried in numpy  and not worth the effort.  You'd have much better luck trying to improve the quality of the simulations on our end.  Most likely this could be solved by having a bit more realistic simulations with a thick Ewald sphere etc.  I think at this point we are maybe borrowing trouble but I don't do these experiments so I might need additional help. 

Edit: Also the new simulation is more inline with ASTAR so maybe just call this a win?

rbjorge:
I thought maybe the old and new method used different default values for the scattering parameters and the shape factor model etc., but I see now that they are the same. With the new simulations, the 011 reflection  is stronger than 010 in this zone axis. In the old simulations, the 010 reflection was stronger than the 011 reflection. This can also be seen in @CSSFrancis's simulations above. Incidentally, ReciPro also gives 011 as stronger, both in the list of structure factors and in the ZAP. It would be interesting to know where the difference originates (presumably in the two different `calculate_ed_data()` functions?), but probably not critical for template matching.

The different crystal reference frames in orix and diffpy are described here: [https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html](https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html)

CSSFrancis:
> I thought maybe the old and new method used different default values for the scattering parameters and the shape factor model etc., but I see now that they are the same. With the new simulations, the 011 reflection is stronger than 010 in this zone axis. In the old simulations, the 010 reflection was stronger than the 011 reflection. This can also be seen in @CSSFrancis's simulations above. Incidentally, ReciPro also gives 011 as stronger, both in the list of structure factors and in the ZAP. It would be interesting to know where the difference originates (presumably in the two different `calculate_ed_data()` functions?), but probably not critical for template matching.
> 
> The different crystal reference frames in orix and diffpy are described here: https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html

@rbjorge I'll take it one step further to say that the new and the old function are actually identical for calculating intensity which is part of the reason that this has been so weird.  The only real difference is:

1. We use orix to determine the reciprocal lattice vectors
2. We use orix to rotate the reciprocal lattice vectors.  

I think that something wasn't being properly handled in relation to one of these two things in the old simulations which is now fixed in the new simulations.  Most likely we weren't handling rotations optimally, but it's possible that we didn't properly account for non-cubic lattices as well.

rbjorge:
> @viljarjf Huh all from a change of:
> 
> `[-4.68628465e-01, -2.98018179e-01, 1.70229751e-16] -->[-4.68628465e-01, -2.98018179e-01, 1.41346131e-16]`
> 
By the way, I see that the greatest difference is in the z-coordinate. It could be significant? Edit: or not...1e-16 reciprocal ångströms is basically zero.

CSSFrancis:
@rbjorge @viljarjf @pc494 how are we feeling about this now?

viljarjf:
I might have found the bug causing the discrepancy: #203

pc494:
Does #205 mean we can close this @CSSFrancis?

CSSFrancis:
Close in favor of #205

PR #205, opened 2024-04-12
Use phase class2

CSSFrancis:
This simplified the commit history for #201 and focuses the scope of the PR.  Hopefully that makes it (slightly) more manageable when checking/merging.

#### Progress of the PR
- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)

```python
>>> from diffsims.utils import sim_utils
>>> # Your new feature...
```

#### For reviewers

      in `.zenodo.json`.


CSSFrancis:
@viljarjf, can you review this? It seems like we should get this in, and then we can work on getting the rest of `diffsims` ready for a release.

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,422 @@
...
+
+        Returns
+        -------
+        diffsims.sims.diffraction_simulation.DiffractionSimulation
+            The data associated with this structure and diffraction setup.
+        """
+        if isinstance(phase, Phase):
+            phase = [phase]
+        if isinstance(rotation, Rotation):
+            rotation = [rotation]
```
viljarjf, line 178: If the lengths of `phase` or `rotation` are different, then the longer one gets silently ignored in the `zip` below. Maybe we should re-use the rotations if there are more phases. If there are more rotations than phases then I'm not so sure what is best.
```suggestion
        if isinstance(phase, Phase):
            phase = [phase]
        if isinstance(rotation, Rotation):
            rotation = [rotation] * len(phase)
```

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,422 @@
...
+                    g_indices[:, 2],
+                ]
+            ).T
+
+        hkls_labels = ["".join([str(int(x)) for x in xs]) for xs in g_indices]
+
+        peaks = {}
+        for l, i, g in zip(hkls_labels, i_hkl, g_hkls):
+            peaks[l] = [i, g]
+
```
viljarjf: This label string might contain duplicates, e.g. (1 2 10) -> "1210" <- (12 1 0), which would override the dictionary value

diffsims/simulations/__init__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017-2023 The diffsims developers
```
viljarjf: Not sure how the copyright is kept up to date, but seeing as this is a new file it should probably be up to date
```suggestion
# Copyright 2017-2024 The diffsims developers
```

diffsims/simulations/simulation2d.py
```diff
@@ -0,0 +1,619 @@
...
+]
+
+
+class PhaseGetter:
+    """A class for getting the phases of a simulation library.
+
+    Parameters
+    ----------
+    phases : Sequence[Phase]
+        The phases in the library.
```
viljarjf:
```suggestion
    simulation: Simulation2D
        The simulation to get from.
```

diffsims/simulations/simulation2d.py
```diff
@@ -0,0 +1,619 @@
...
+        )
+
+
+class RotationGetter:
+    """A class for getting a Rotation of a simulation library.
+
+    Parameters
+    ----------
+    phases : Sequence[Phase]
+        The phases in the library.
```
viljarjf:
```suggestion
    simulation: Simulation2D
        The simulation to get from.
```

diffsims/simulations/simulation2d.py
```diff
@@ -0,0 +1,619 @@
...
+        if self.has_multiple_phases:
+            return copy.deepcopy(
+                self.coordinates[self.phase_index][self.rotation_index]
+            )
+        elif not self.has_multiple_phases and self.has_multiple_rotations:
+            return copy.deepcopy(self.coordinates[self.rotation_index])
+        else:
+            return copy.deepcopy(self.coordinates)
+
+    def get_current_rotation(self):
```
viljarjf: Bit of a misleading function name, since the rotations are normally `Rotation` objects, but the docstring clarifies.

diffsims/simulations/simulation2d.py
```diff
@@ -0,0 +1,619 @@
...
+        """Returns the matrix for the current matrix"""
```
viljarjf:
```suggestion
        """Returns the matrix for the current rotation"""
```

diffsims/tests/generators/test_simulation_generator.py
```diff
@@ -0,0 +1,262 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017-2023 The diffsims developers
```
viljarjf:
```suggestion
# Copyright 2017-2024 The diffsims developers
```

diffsims/utils/shape_factor_models.py
```diff
@@ -217,3 +218,53 @@ def lorentzian_precession(
...
+    for i, (excitation_error_i, r_spot_i) in enumerate(zip(excitation_error, r_spot)):
+
+        def integrand(theta):
+            # Equation 8 in L.Palatinus et al. Acta Cryst. (2019) B75, 512-522
+            S_zero = excitation_error_i
+            variable_term = r_spot_i * (phi) * np.cos(theta)
+            return shape_function(S_zero + variable_term, max_excitation, **kwargs)
+
+        # average factor integrated over the full revolution of the beam
+        shf[i] = (1 / (2 * np.pi)) * quad(integrand, 0, 2 * np.pi)[0]
```
viljarjf, line 269: This can be vectorized with `quad_vec`, but that can wait

setup.py
```diff
@@ -78,7 +78,7 @@
         "matplotlib         >= 3.3",
         "numba",
         "numpy              >= 1.17.3",
-        "orix               >= 0.9",
+        "orix               >= 0.11",
```
viljarjf: Should this not be 0.12.1 now? Or is dependency updating perhaps done separately?

viljarjf (commented):
Looks good! See comments for some minor things. I did not go through all the tests very thoroughly, but they all pass at least.

CSSFrancis:
@viljarjf This should be better now based on your comments. Thanks, let me know what you think and maybe we can merge this?

diffsims/generators/simulation_generator.py
CSSFrancis: I just raised an error when this is the case. I think that is the best way to handle this without making it too complicated.

setup.py
CSSFrancis: Yes :)

diffsims/utils/shape_factor_models.py
CSSFrancis: Let's make an issue for this and we can handle this later.

diffsims/simulations/simulation2d.py
CSSFrancis: Done!

viljarjf (approved):
I think this can be merged :)

hakonanes:
I'd like to review it given it touches on `ReciprocalLatticeVector`, which is heavily used in kikuchipy workflows. I'm not so sure about the `intensity` parameter.

hakonanes:
I can review within the next few days.

hakonanes:
So, sorry about the hold-up, @CSSFrancis.

Couple of questions right away:
1. Could you explain the reason for the `intensity` attribute in `ReciprocalLatticeVector`? If you're interested in the kinematical scattering intensity I = |F|^2, this can be calculated from `F = calculate_structure_factor()`. The new attribute lacks a docstring. To me, just looking at the new additions and tests, the attribute looks like a placeholder for any array you want, attached to the `ReciprocalLatticeVector` class.
2. Why cannot the new simulation classes be part of the current simulation module, `diffsims.sims`? Having two modules for diffraction simulations is confusing.

CSSFrancis:
> So, sorry about the hold-up, @CSSFrancis.
> 
> Couple of questions right away:
> 
> 1. Could you explain the reason for the `intensity` attribute in `ReciprocalLatticeVector`? If you're interested in the kinematical scattering intensity I = |F|^2, this can be calculated from `F = calculate_structure_factor()`. The new attribute lacks a docstring. To me, just looking at the new additions and tests, the attribute looks like a placeholder for any array you want, attached to the `ReciprocalLatticeVector` class.

That's a good question; it's mostly that we __really__ want to calculate things once and then not do it again, or it's going to really slow down the orientation mapping. We can add a function `calculate_scattering_intensity` to the DiffractionVectors class? I can refactor that. That data also contains information about the intersection of the vectors and the Ewald sphere.  Maybe this should be a separate object but it needs to be stored somewhere.

> 2. Why cannot the new simulation classes be part of the current simulation module, `diffsims.sims`? Having two modules for diffraction simulations is confusing.

Shorting names is fairly un-pythonic. A 1-1 mapping of Signal1D -> Simulation1D and Signal2D --> Simulation2D makes more sense in the long term.  I want to make this change long-term in diffsims and don't want to break the pyxem API when that happens.  I'm open to suggestions.

CSSFrancis:
> > So, sorry about the hold-up, @CSSFrancis.
> > Couple of questions right away:
> > 
> > 1. Could you explain the reason for the `intensity` attribute in `ReciprocalLatticeVector`? If you're interested in the kinematical scattering intensity I = |F|^2, this can be calculated from `F = calculate_structure_factor()`. The new attribute lacks a docstring. To me, just looking at the new additions and tests, the attribute looks like a placeholder for any array you want, attached to the `ReciprocalLatticeVector` class.
> 
> That's a good question; it's mostly that we **really** want to calculate things once and then not do it again, or it's going to really slow down the orientation mapping. We can add a function `calculate_scattering_intensity` to the DiffractionVectors class? I can refactor that.  We also need to know information about the Ewald sphere to calculate and determine which vectors intersect with the sphere.  Maybe this should be a separate array/obejct, but it needs to be stored somewhere.
> 
> > 2. Why cannot the new simulation classes be part of the current simulation module, `diffsims.sims`? Having two modules for diffraction simulations is confusing.
> 
> Shorting names is fairly un-pythonic. A 1-1 mapping of Signal1D -> Simulation1D and Signal2D --> Simulation2D makes more sense in the long term. I want to make this change long-term in diffsims and don't want to break the pyxem API when that happens. I'm open to suggestions.

hakonanes:
> we really want to calculate things once and then not do it again

I totally understand that. That's why e.g. the `structure_factor` attribute carries over when slicing `ReciprocalLatticeVector` already.

> We can add a function calculate_scattering_intensity to the DiffractionVectors class?

Hm, where is this class? We have the `calculate_structure_factor()` method to calculate kinematical scattering intensities for `ReciprocalLatticeVector`. This has served us well in kikuchipy for calculating kinematical EBSD patterns. We should ideally have one way of doing things. Having attributes for the structure factor `F`, calculated from `calculate_structure_factor()`, and then for an intensity `I = |F|^2`, which may not be derived from `F`, is confusing (to me).

> That data also contains information about the intersection of the vectors and the Ewald sphere. Maybe this should be a separate object but it needs to be stored somewhere.

But, the end result is ultimately structure factors? Or the intensities directly? And what is the current route for calculating these structure factors (steps and input parameters)? It may be best to include that route into the `ReciprocalLatticeVector` class.

>  I want to make this change long-term in diffsims and don't want to break the pyxem API when that happens. I'm open to suggestions.

This is fine, and something I agree with in general. Can I suggest to present this idea in an issue for a discussion, albeit a brief one?

CSSFrancis:
> I totally understand that. That's why e.g. the `structure_factor` attribute carries over when slicing `ReciprocalLatticeVector` already.
> 
> > We can add a function calculate_scattering_intensity to the DiffractionVectors class?
> 
> Hm, where is this class? We have the `calculate_structure_factor()` method to calculate kinematical scattering intensities for `ReciprocalLatticeVector`. This has served us well in kikuchipy for calculating kinematical EBSD patterns. We should ideally have one way of doing things. Having attributes for the structure factor `F`, calculated from `calculate_structure_factor()`, and then for an intensity `I = |F|^2`, which may not be derived from `F`, is confusing (to me).

I misspoke sorry :) Trying to deal with too many things at one time... I meant  `ReciprocalLatticeVector` (`DiffractionVectors` is a similar class in pyxem). 

Maybe it's best if we think about the entire workflow and determine which parts should be methods for `ReciprocalLatticeVector`

We want to:

1 - Copy the `ReciprocalLatticeVector` class 
2 - Rotate the `ReciprocalLatticeVector` (to do this I added the `rotate_from_matrix` function
3 - Determine which `ReciprocalLatticeVector` vectors intersect the Ewald Sphere 
4 - Calculate the diffraction intensities. This includes the Debye Waller factors __and__ the shape function.
5 - Remove `ReciprocalLatticeVector` vectors below a cutoff intensity

For step 4, we can derive the intensities from the structure factor. We just need to make sure we are __only__ calculating this once; otherwise, it might slow down the template matching. We must also add Debye Waller factors and the shape function.   Having these factors in the `ReciprocalLatticeVector` class is a bit weird as they are more closely related to the  SimulationGenerator.

> > That data also contains information about the intersection of the vectors and the Ewald sphere. Maybe this should be a separate object but it needs to be stored somewhere.
> 
> But, the end result is ultimately structure factors? Or the intensities directly? And what is the current route for calculating these structure factors (steps and input parameters)? It may be best to include that route into the `ReciprocalLatticeVector` class.

See above. The result that we care about if the diffraction intensity for comparison.  I don't think we want those steps in the `ReciprocalLatticeVector` as it requires that the `ReciprocalLatticeVector` object has information about the Ewald sphere which seems like it should be part of the SimulationGenerator.  I could be convinced otherwise.

I guess the question is: If you wanted to plot the diffraction spots __and__ the Kikuchi lines, how would that look/ how is that handled?  It might be nice to do something like: `s.plot(diffraction=True, kikuchi_lines=True)`
> > I want to make this change long-term in diffsims and don't want to break the pyxem API when that happens. I'm open to suggestions.
> 
> This is fine, and something I agree with in general. Can I suggest to present this idea in an issue for a discussion, albeit a brief one?

Sounds good :)

hakonanes:
> I don't think we want those steps in the ReciprocalLatticeVector as it requires that the ReciprocalLatticeVector object has information about the Ewald sphere which seems like it should be part of the SimulationGenerator.

I agree! It sounds like we want the generator to *use* `ReciprocalLatticeVector` (composition), rather than adding to the vector class. I'm of the opinion that no microscope- or geometry-specific parameters should be part of this class.

The only method that takes a microscope parameter, an accelerating voltage, is [`calculate_theta()`](https://diffsims.readthedocs.io/en/latest/reference/generated/diffsims.crystallography.ReciprocalLatticeVector.calculate_theta.html#diffsims.crystallography.ReciprocalLatticeVector.calculate_theta) for populating the `theta` attribute, twice the Bragg angle. This should arguably not be an attribute of the class, because information of the voltage is lost upon return. And having a `ReciprocalLatticeVector.voltage` attribute makes no sense. But, it's been a part of the API for many years, so it should be kept.

> We must also add Debye Waller factors

These are already part of the [diffpy.structure atoms API](https://www.diffpy.org/diffpy.structure/mod_atom.html#diffpy.structure.atom.Atom.Bisoequiv) in the `Atom.Bisoequiv` attribute. Can you use this? The factors are used when calculating the atomic scattering factors, f (used in `ReciprocalLatticeVector.calculate_structure_factor()`):

https://github.com/pyxem/diffsims/blob/bb359b9a6326f22ab414665b005d3e04e5d56340/diffsims/structure_factor/atomic_scattering_factor.py#L62-L64

> It might be nice to do something like: s.plot(diffraction=True, kikuchi_lines=True)

What is `s` here? What information about the microscope, phase, detector, and scattering vectors does it have?

We have plotting of Kikuchi lines in kikuchipy (see [this tutorial](https://kikuchipy.org/en/stable/tutorials/geometrical_ebsd_simulations.html#Create-simulations-on-detector)). It may be that the simulation generator should be upstreamed to diffsims, but that would also require upstreaming `EBSDDetector`, which powers most of kikuchipy... Perhaps just taking necessary code from kikuchipy is fine for the Kikuchi lines. I don't know.

CSSFrancis:
> I agree! It sounds like we want the generator to _use_ `ReciprocalLatticeVector` (composition), rather than adding to the vector class. I'm of the opinion that no microscope- or geometry-specific parameters should be part of this class.
> 
> The only method that takes a microscope parameter, an accelerating voltage, is [`calculate_theta()`](https://diffsims.readthedocs.io/en/latest/reference/generated/diffsims.crystallography.ReciprocalLatticeVector.calculate_theta.html#diffsims.crystallography.ReciprocalLatticeVector.calculate_theta) for populating the `theta` attribute, twice the Bragg angle. This should arguably not be an attribute of the class, because information of the voltage is lost upon return. And having a `ReciprocalLatticeVector.voltage` attribute makes no sense. But, it's been a part of the API for many years, so it should be kept.

@hakonanes Okay, so what do we do about the diffraction intensity, which depends on the microscope parameters?  I can separate them if you like and make that an attribute of Simulation2D. Still, then there is always the possibility that you have more intensity values than `ReciprocalLatticeVector`s if someone slices either of them.

> > We must also add Debye Waller factors
> 
> These are already part of the [diffpy.structure atoms API](https://www.diffpy.org/diffpy.structure/mod_atom.html#diffpy.structure.atom.Atom.Bisoequiv) in the `Atom.Bisoequiv` attribute. Can you use this? The factors are used when calculating the atomic scattering factors, f (used in `ReciprocalLatticeVector.calculate_structure_factor()`):
>
> https://github.com/pyxem/diffsims/blob/bb359b9a6326f22ab414665b005d3e04e5d56340/diffsims/structure_factor/atomic_scattering_factor.py#L62-L64

I'm not really trying to change everything all at once... Can we leave that to another PR? This PR is just supposed to use the orix phase class and `ReciprocalLatticeVector` so we should focus on making that one change and __then__ clean up any other part that we feel isn't working.  This PR is holding up __a lot__ and we need to figure out how to get this in.

> > It might be nice to do something like: s.plot(diffraction=True, kikuchi_lines=True)
> 
> What is `s` here? What information about the microscope, phase, detector, and scattering vectors does it have?
> 

That would be your Simulation2D object. It might make more sense to calculate the two separately and add them to the same plot. It might also be nice to have one SimulationGenerator object that can do both kikuchi and vector diffraction.

> We have plotting of Kikuchi lines in kikuchipy (see [this tutorial](https://kikuchipy.org/en/stable/tutorials/geometrical_ebsd_simulations.html#Create-simulations-on-detector)). It may be that the simulation generator should be upstreamed to diffsims, but that would also require upstreaming `EBSDDetector`, which powers most of kikuchipy... Perhaps just taking necessary code from kikuchipy is fine for the Kikuchi lines. I don't know.

For now maybe we can just have a tutorial that makes a combined plot is someone is interested.

CSSFrancis:
@hakonanes I guess the question is what needs to be changed for this to be merged?  Most of this is just copied from the old function.  It's not really anything new, just makes it significantly faster and (somewhat) cleaner.

hakonanes:
>  I can separate them if you like and make that an attribute of Simulation2D

That sounds like the best approach. I hope you can find a clean way of combining the intensity and vectors. As I've argued above, I don't think the intensity should ~not~ be part of the reciprocal lattice vector class. I might be convinced otherwise by arguments relating to the nature of the vectors and the intensities.

> Still, then there is always the possibility that you have more intensity values than ReciprocalLatticeVectors if someone slices either of them.

Can you add a slicer to the simulator instead? You can then slice the intensity array and vectors accordingly.

> Can we leave that to another PR?

Absolutely! I just pointed out that you may not need to have separate arrays for the Debye-Waller factors, and just use diffpy.structure instead.

> This PR is just supposed to use the orix phase class and ReciprocalLatticeVector so we should focus on making that one change

I realize that, and I'm sorry for holding this up. But I'm not a fan of adding use-case-specific attributes and function parameters to a general class that is instrumental to workflows in downstream packages, like e.g. kikuchipy.

> I guess the question is what needs to be changed for this to be merged?

I suggest to remove the new additions to `ReciprocalLatticeVector` and instead use the vectors alongside the intensities array in a simulator. I can review after that is done.

CSSFrancis:
@hakonanes, how do you feel about a subclass of `ReciporicalLatticeVector`, something like `DiffractingVector`, which specifically handles  `ReciporicalLatticeVector`s that intersect with the Ewald Sphere?

hakonanes:
That doesn't sound too bad. If we subclass, and don't use the structure factor calculations already there, these should be overridden to return not implemented, though.

CSSFrancis:
@hakonanes Can you look over this when you get the chance?

hakonanes:
Yes, sorry, will do within the next few days!

pc494 (commented):
I'll be deferring to @hakonanes here.

CSSFrancis:
> Yes, sorry, will do within the next few days!

@hakonanes Any chance we can get this done this week?  It would be good to have this done and release a new pyxem version before the Diamond workshop on the 13th.

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -1070,7 +1076,7 @@ def from_highest_hkl(cls, phase, hkl):
         return cls(phase, hkl=idx).unique()
 
     @classmethod
-    def from_min_dspacing(cls, phase, min_dspacing=0.7):
+    def from_min_dspacing(cls, phase, min_dspacing=0.7, include_zero_beam=False):
```
hakonanes: "Zero beam" is TEM-diffraction specific. Can we rename this to `include_zero_vector`?

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -1083,6 +1089,8 @@ def from_min_dspacing(cls, phase, min_dspacing=0.7):
             Smallest interplanar spacing to consider. Default is 0.7,
             in the unit used to define the lattice parameters in
             ``phase``, which is assumed to be Ångström.
+        include_zero_beam: bool
+            If ``True``, include the zero beam in the set of vectors.
```
hakonanes: Something along the lines of "Whether to include the zero vector (000). Default is False".

diffsims/crystallography/reciprocal_lattice_vector.py
hakonanes: And, I think we should add this to all class methods.

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -502,6 +503,11 @@ def scattering_parameter(self):
 
         return 0.5 * self.gspacing
 
+    def rotate_from_matrix(self, rotation_matrix):
```
hakonanes: Where is this used or going to be used? If internally, I'd suggest to remove it and do the computation where necessary.

How is it going to be used? There is no docstring. The requirement is a (3, 3) matrix, this should be stated.

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -502,6 +503,11 @@ def scattering_parameter(self):
...
+        return ReciprocalLatticeVector(
+            phase=self.phase, xyz=np.matmul(rotation_matrix.T, self.data.T).T
```
hakonanes: This should be `np.matmul(self.data, rotation_matrix.T)`, no?

That is in line with orix. Compare e.g.

```python
import numpy as np
from diffsims.crystallography import ReciprocalLatticeVector


phase = Phase(point_group="m-3m")
g = ReciprocalLatticeVector.from_highest_hkl(phase, [1, 1, 1])
g_m = g.to_miller()

R = Rotation.random()
om = R.to_matrix().squeeze()

g1 = R * g_m
g2 = ReciprocalLatticeVector(
    phase=g.phase, xyz=np.matmul(g.data, om.T)
)

np.allclose(g1.data, g2.data)
# True
```

diffsims/tests/crystallography/test_reciprocal_lattice_vector.py
```diff
@@ -108,6 +108,14 @@ def test_repr(self, ferrite_phase):
             "[[ 1.  1.  0.]",
         ]
 
+    @pytest.mark.parametrize("degrees", [True, False])
```
hakonanes: `to_polar()` is inherited from `Vector3d` in orix. I see no need to test it here, so I suggest to remove it.

diffsims/crystallography/diffracting_vector.py
```diff
@@ -0,0 +1,116 @@
...
+                [
+                    value,
+                ]
+                * self.size
+            )
+        if len(value) != self.size:
+            raise ValueError("Length of intensity array must match number of vectors")
+        self._intensity = np.array(value)
+
+    def calculate_structure_factor(self):
```
hakonanes: We should improve the [technical reference](https://diffsims--205.org.readthedocs.build/en/205/reference/generated/diffsims.crystallography.DiffractingVector.html) (API docs) of this class. E.g., `intensity` needs a docstring.

And the only method listed is `calculate_structure_factor()`, which is not supported. It would be good to do [as in orix](https://github.com/pyxem/orix/blob/a1be2697dc0cf02974b00d06d94272de77efeafa/doc/conf.py#L345-L364) and include inherited members within the package only. We may actually try including inherited members from orix as well.

What do you think?

diffsims/crystallography/diffracting_vector.py
```diff
@@ -0,0 +1,116 @@
...
+        raise NotImplementedError(
+            "Structure factor calculation not implemented for DiffractingVector"
```
hakonanes: Perhaps "Structure factor calculation not implemented for DiffractionVector. Use ReciprocalLatticeVector instead."?

hakonanes (changes requested):
Here're some suggestions for improvement. I'd very much like your thoughts on my points, @CSSFrancis. Note that I've only looked at `ReciprocalLatticeVector`.

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: @hakonanes This kind of has to be included. Otherwise, there isn't really a reason to use the RLV.  We need to rotate the vectors and then determine the intersection with the Ewald sphere. 

I'm open to suggestions...

We need a method that basically goes from RLV --> Rotate --> Select --> Create DiffractingVector

diffsims/crystallography/diffracting_vector.py
CSSFrancis: That sounds like a good idea.  I can try to make a follow up PR for that.

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: This changes the simulation results quite a bit.  I don't think we want to do this unless we have a better idea of tracking what this change does and how we can retain some consistency with the simulations.

diffsims/tests/crystallography/test_reciprocal_lattice_vector.py
CSSFrancis: Done!

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: I added a test to check that the results remain similar. Maybe @viljarjf you have a comment on this

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: Sounds good.  I added this to the `from_highest_hkl` method as well.

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: There is a bit of a problem where once you rotate the diffraction vector the hkl no longer make sense. 

I'm not sure the proper solution though likely we would need to rethink the Miller/ReciporicalLatticeVector class.

diffsims/crystallography/reciprocal_lattice_vector.py
viljarjf: I am a fan of staying consistent with orix, at least. Preferably, the ReciprocalLatticeVector should be rotated by multiplying with a Rotation, rather than a rotation matrix, but this currently does not work as reported in #211.
Transposing the rotation matrix inverts the rotation, which might be why the simulations change.
As for the test you added, I believe the .npy-file is missing right now?

CSSFrancis: @viljarjf i think the file is there but it's being excluded in the build? @hakonanes Might have a better idea

CSSFrancis:
@hakonanes Okay so I think we have to sort out some things. 

1. Can the `RLV` be ragged?  --> Probably not unless the Miller and Vector3D class supports it.
2. Should we allow the `RLV` to be rotated?  ~~I think the answer to this is yes. But this requires that we add a `.rotation` attribute to the `RLV` so that the `.hkl` value is computed like: https://github.com/CSSFrancis/diffsims/blob/9bf940441bff960c8b11642b26cf79d66f13c8a0/diffsims/simulations/simulation2d.py#L556C1-L560C18~~ --> Lets just keep this so we can't rotate it just to get this in.
3. Should the `RLV` have extra information like the intensity? --> No.
5. How do we store only the selected (ragged) diffracting RLV.  --> This can just be an array of diffracting vectors.

CSSFrancis:
Okay, for the sake of getting this in and because we really, really need to get this out before the Diamond workshop in one week, I'm just going to make the DiffractingVector class private. We are getting super stuck trying to make `diffsims` perfect rather than just making it better... At the end of the day it's a 0.6.0 version, it's not a major release, and we aren't breaking anything.


We can make this a beta feature or something like that if it makes you feel better.

viljarjf:
If the existence of `rotate_from_matrix`-method is what is blocking this, and it is only used in `SimulationGenerator.get_intersecting_reflections`, then how about this instead:
```python
    def get_intersecting_reflections(
        self,
        recip: ReciprocalLatticeVector,
        rot: Rotation,
        wavelength: float,
        max_excitation_error: float,
        shape_factor_width: float = None,
        with_direct_beam: bool = True,
    ):
        ...
        initial_hkl = recip.hkl
        # To ease calculations, we transform the reciprocal lattice vectors rather than the electron beam.
        # Therefore, we need to invert the rotation, as it normally transforms the beam (lab2crystal).
        rotated_vectors = (~rot * recip.to_miller()).data

        if with_direct_beam:
            rotated_vectors = np.vstack([rotated_vectors, [0, 0, 0]])
            initial_hkl = np.vstack([initial_hkl, [0, 0, 0]])

        # Identify the excitation errors of all points (distance from point to Ewald sphere)
        ...
```
The behaviour is the exact same as before.
I can commit this change if you want @CSSFrancis?

CSSFrancis:
@viljarjf sure you can make a PR to my branch and I can merge it if you want :)

I think that is the right way to do things as the hkl complicates things.

I still think the DiffractingVector class being private might be a good thing. It's only really used for simulation and we need to figure out the best way to handle the hkl. Based on the timeline minimal changes but still being able to do everything we want is probably the best way forward.

CSSFrancis:
Also just wanted to say thank you @viljarjf and @hakonanes for your help and comments. Sorry if I've been pushing this a bit too much :)

CSSFrancis:
Okay, so I made a couple of changes:

1. Fixed the failing test.  I forgot that the dir is set to the top-level pytest function. 
2. Removed the `rotate_from_matrix` method and used the `Rotation` class base on @viljarjf 
3. Make the DiffractingVector class private.  (we could probably make it public.)
4. Added a rotation attribute to the `DiffractingVector` class to handle hkl.

diffsims/simulations/simulation2d.py
```diff
@@ -0,0 +1,746 @@
...
+        """
+
+        flattened_vectors = [sim for sim in self]
+        max_num_spots = max([v.size for v in flattened_vectors])
+
+        r_templates = np.zeros((len(flattened_vectors), max_num_spots))
+        theta_templates = np.zeros((len(flattened_vectors), max_num_spots))
+        intensities_templates = np.zeros((len(flattened_vectors), max_num_spots))
+        for i, v in enumerate(flattened_vectors):
+            r, t, _ = v.to_polar()
```
viljarjf: I am getting some strange results when I try to template match, and I believe this line is the cause. Should it not be `t, _, r = v.to_polar()`? Seeing as `Vector3d.to_polar` returns (azimuthal, polar, radius) spherical coordinates. By inspection, the current way yields `t`-values close to pi/2, which would make sense for the polar angle with a mostly flat Ewald's sphere.

CSSFrancis: @viljarjf Sorry, I've gotten kind of confused with all of these comments and changes.  We don't want to use that function at all... We want the radius of the projection of the vector onto the x,y plane.

CSSFrancis: @viljarjf This (should) be better

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class DiffractingVector(ReciprocalLatticeVector):
+    r"""Reciprocal lattice vectors :math:`(hkl)` for use in electron
+    diffraction analysis and simulation.
+
+    All lengths are assumed to be given in Å or inverse Å.
+
+    This extends the :class:`ReciprocalLatticeVector` class.  Diffracting Vectors
+    focus on the subset of reciporical lattice vectors that are relevant for
```
hakonanes:
```suggestion
    This extends the :class:`ReciprocalLatticeVector` class.  `DiffractingVector`
    focus on the subset of reciprocal lattice vectors that are relevant for
```

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,153 @@
...
+    electron diffraction based on the intersection of the Ewald sphere with the
+    reciprocal lattice.
+
+    This class is only used internally to store the DiffractionVectors generated from the
+    DiffractionSimulation class. It is not (currently) intended to be used directly by the user.
```
hakonanes:
```suggestion
    This class is only used internally to store the diffracting vectors generated from a
    :class:`~diffsims.simulations.DiffractionSimulation`. It is not intended to be used
    directly by the user.
```

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,153 @@
...
+    hkl : numpy.ndarray, list, or tuple, optional
+        Indices of reciprocal lattice vector(s). Default is ``None``.
+        This, ``xyz``, or ``hkil`` is required.
+    hkil : numpy.ndarray, list, or tuple, optional
+        Indices of reciprocal lattice vector(s), often preferred over
+        ``hkl`` in trigonal and hexagonal lattices. Default is ``None``.
+        This, ``xyz``, or ``hkl`` is required.
+    intensity : numpy.ndarray, list, or tuple, optional
+        Intensity of the diffraction vector(s). Default is ``None``.
+    rotation : orix.quaternion.Rotation, optional
```
hakonanes, line 57: I don't understand why the vectors need to know which orientation took them to where they are now. Can you explain?

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,153 @@
...
+                    value,
+                ]
+                * self.size
+            )
+        if len(value) != self.size:
+            raise ValueError("Length of intensity array must match number of vectors")
+        self._intensity = np.array(value)
+
+    @property
+    def lattice_aligned_data(self):
```
hakonanes: So these vectors are the same as they were before rotated by a rotation. Can't we *separately* keep track of a set of starting vectors, a rotation, and a set of rotated vectors? Why do these separate things have to be stored in a single class instance?

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def calculate_structure_factor(self):
+        raise NotImplementedError(
+            "Structure factor calculation not implemented for DiffractionVector. "
+            "Use ReciprocalLatticeVector instead."
+        )
+
+    def to_flat_polar(self):
+        """Return the vectors in polar coordinates as projected onto the x,y plane"""
+        r = np.linalg.norm(self.data[:, :2], axis=1)
```
hakonanes: What if `self.data.ndim > 2`? The reciprocal lattice vectors, as normal 3D vectors, are allowed to have a "navigation shape" > 1. I suggest to flatten before extracting the x, y coordinates.

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -18,6 +18,7 @@
 
 from collections import defaultdict
 from copy import deepcopy
+from typing import Tuple
```
hakonanes: This line can be dropped.

diffsims/generators/simulation_generator.py
```diff
@@ -0,0 +1,428 @@
...
+"""Kinematic Diffraction Simulation Generator."""
+
+from typing import Union, Sequence
+import numpy as np
+
+from orix.quaternion import Rotation
+from orix.crystal_map import Phase
+
+from diffsims.crystallography import ReciprocalLatticeVector, DiffractingVector
```
hakonanes: Can the diffracting vector class be private?

You state in its docstring that it is not meant to be used by end-users. This tells me that the logic should be a simpler class or sets of functions to handle this particular use-case, instead of a public class with many caveats.

hakonanes: If the user wants the original reciprocal lattice vectors, rotations, or rotated vectors, can they instead get `ReciprocalLatticeVector` or `Rotation` (preferably the more powerful `Orientation`)?

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -1067,10 +1070,12 @@ def from_highest_hkl(cls, phase, hkl):
         """
 
         idx = _get_indices_from_highest(highest_indices=hkl)
+        if include_zero_vector:
```
hakonanes, line 1072: Can we add a test for this change?

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -1128,6 +1135,8 @@ def from_min_dspacing(cls, phase, min_dspacing=0.7):
         dspacing = 1 / phase.structure.lattice.rnorm(hkl)
         idx = dspacing >= min_dspacing
         hkl = hkl[idx]
+        if include_zero_vector:
```
hakonanes, line 1139: Can we add a test for this change?

hakonanes (changes requested):
I've had a look at the new changes.

diffsims/simulations/__init__.py
hakonanes: We use the licenseheaders package to automate this: https://pypi.org/project/licenseheaders/.

diffsims/crystallography/reciprocal_lattice_vector.py
hakonanes: Thanks for making the change!

diffsims/crystallography/_diffracting_vector.py
CSSFrancis: I think I was trying a bit too hard to follow the `RLV`. The problem is that once you apply a rotation to a RLV, the hkl no longer makes sense. In essence, you rotate the entire Lattice object, which we could do, but if we want to be able to multiple by a `Rotation` object we have to change how we think about that.  (See #211)

diffsims/crystallography/_diffracting_vector.py
CSSFrancis: Hmm.... I think this is related to above.  What use is the RLV if it __doesn't__ have information about the Rotation otherwise it is just the same as a `Vector3D` with confusing hkl values that don't really make physical sense and look correct because they are rounded.

diffsims/crystallography/_diffracting_vector.py
CSSFrancis: Good point.

diffsims/crystallography/reciprocal_lattice_vector.py
CSSFrancis: Yeah, I was kind of surprised that didn't cause an error with the test suite.

diffsims/crystallography/reciprocal_lattice_vector.py
hakonanes: Now that I think about it, it might be fine that reciprocal lattice vectors cannot be rotated... If that functionality is needed, one can just use regular Miller (hkl) vectors instead. We can continue discussing in #211.

CSSFrancis:
@hakonanes I think lots of the confusion is related to not properly handling the rotation with the RLV.

We __need__ to track the rotations applied to the RLV to recover the hkl.  This is possible with https://github.com/pyxem/orix/pull/499 and adding the `rotation` property to the RLV.

```python
from diffsims.crystallography import ReciprocalLatticeVector
import numpy as np

phase = Phase(point_group="m-3m")
g = ReciprocalLatticeVector(phase, [[1, 1, 1], [1,0,0]])

R1 = Rotation.random(2)
random1 = R1 * g
R2 = Rotation.random(2)
random2 =R2*random1

assert np.allclose(g.hkl, random1.hkl) # Previously this would fail
assert np.allclose(g.hkl, random2.hkl) # Previously this would fail
```

I can push this change if you want but it requires https://github.com/pyxem/orix/pull/499.

I think this helps the logic quite a bit but does change the ReciprocalLatticeVector class a little bit.

setup.cfg
```diff
@@ -28,4 +28,5 @@ known_excludes =
      .git/**
      doc/build/**
      htmlcov/**
-     *.code-workspace
\ No newline at end of file
+     *.code-workspace
+     **/*.npy
```
hakonanes, line 32: See #212.

hakonanes:
> I'm just going to make the DiffractingVector class private

Perfect. Then I've got no objections to how that class is used internally in diffsims.

diffsims/crystallography/__init__.py
```diff
@@ -27,11 +27,13 @@
...
 from diffsims.crystallography.reciprocal_lattice_vector import ReciprocalLatticeVector
+from diffsims.crystallography._diffracting_vector import DiffractingVector
 
 __all__ = [
     "get_equivalent_hkl",
     "get_highest_hkl",
     "get_hkl",
     "ReciprocalLatticePoint",
     "ReciprocalLatticeVector",
+    "DiffractingVector",
```
hakonanes: This should be removed so that it is not included in the public API, which it currently is: https://diffsims--205.org.readthedocs.build/en/205/reference/generated/diffsims.crystallography.html.

CSSFrancis: Done

CSSFrancis:
@hakonanes I made some changes based on @viljarjf's suggestion in #211

diffsims/crystallography/reciprocal_lattice_vector.py
```diff
@@ -150,6 +153,53 @@ def __repr__(self):
...
+            if value.size != 1:
+                raise ValueError("Rotation must be a single rotation")
+            matrix = value.to_matrix().squeeze()
+        elif isinstance(value, np.ndarray):
+            if value.shape != (3, 3):
+                raise ValueError("Rotation matrix must be 3x3")
+            matrix = value
+        else:
+            raise ValueError("Rotation must be a Rotation or a 3x3 numpy array")
+        self.phase.structure.lattice.setLatPar(baserot=matrix)
```
viljarjf: Phases are passed by refrence, so this would affect all other vectors with the same phase.
This is why I copied the phase in `rotate_with_basis`.
```suggestion
        self.phase = self.phase.deepcopy()
        self.phase.structure.lattice.setLatPar(baserot=matrix)
```
The base rotation is not always the identity as well, especially for hexagonal crystals from cif or initialized from a structure. This comes from the enforced x || a, z || c* convention used in orix (https://orix.readthedocs.io/en/stable/tutorials/crystal_reference_frame.html).
What do you intend the `basis_rotation` member to be used for?

CSSFrancis: Fair point. I'll probably just remove this then. I added it for completeness sake but didn't realize that the base rot was used for hexagonal phases.

CSSFrancis:
Based on:

- https://github.com/pyxem/orix/issues/501
- #211 

I just made everything part of the private DiffractingVectors class.  That way, we can merge things and start working on the next part, the Simulation Part of this code is much more straight forward and I'm much more confident on that implementation. 

@hakonanes can you take another pass through and then we can think about merging?

diffsims/crystallography/_diffracting_vector.py
hakonanes: Thank you for explaining. Given this class is private, I'm OK this design, although we have taken a different approach in kikuchipy.

I think you're here at the heart of what we've discussed in this PR (although I didn't know at first!). In kikuchipy, again, we've avoided this problem entirely by composition. A simulation object stores the perfect, pristine, and un-rotated reciprocal lattice vectors (hkl), the crystal rotations, and the corresponding reflections *on the detector*, i.e. *not by themselves inside a rotated crystal*. Each reflection on the detector points back to one of the un-rotated reciprocal lattice vectors.

diffsims/crystallography/_diffracting_vector.py
```diff
@@ -0,0 +1,181 @@
...
+    def rotate_with_basis(self, rotation):
+        """Rotate both vectors and the basis with a given `Rotation`.
+        This differs from simply multiplying with a `Rotation`,
+        as that would NOT update the basis.
+
+        Parameters
+        ----------
+        rot : orix.quaternion.Rotation
+            A rotation to apply to vectors and the basis.
+        """
```
hakonanes, line 149: I would suggest, and highly appreciate, and explanation that rotating the lattice may violate orix' assumption of crystal axes alignment (e.g. in a *Notes* section). This may lead to undefined behavior when applying the symmetry operations of the phase in the returned vectors.

Also, a description of the returned vectors in a dedicated section would be preferably.

hakonanes: Not for end-users, but for the maintainers of diffsims.

diffsims/tests/crystallography/test_reciprocal_lattice_vector.py
```diff
@@ -20,6 +20,8 @@
 import numpy as np
 from orix.crystal_map import Phase
 from orix.vector import Miller, Vector3d
+from orix.quaternion import Rotation
```
hakonanes: This import is now unnecessary.

diffsims/tests/crystallography/test_reciprocal_lattice_vector.py
```diff
@@ -72,12 +74,20 @@ def test_init_raises(self, nickel_phase):
         with pytest.raises(ValueError, match="Exactly one of "):
             _ = ReciprocalLatticeVector(nickel_phase)
 
+    @pytest.mark.parametrize("include_zero_beam", [True, False])
```
hakonanes: Renaming the variable to `include_vector_vector` would be in line with the parameter name, and would be preferable (in both tests).

hakonanes (changes requested):
Just a few suggestions left. After those are addressed, I'm OK with the parts of this PR that I've reviewed. Will have a final look after that.

CSSFrancis:
> Just a few suggestions left. After those are addressed, I'm OK with the parts of this PR that I've reviewed. Will have a final look after that.

Done!

hakonanes (approved):

CSSFrancis:
@hakonanes Did you want to do another pass through or do you want to merge?

hakonanes:
Fingers crossed the simulations work as you intended in the wild next week!

CSSFrancis:
> Fingers crossed the simulations work as you intended in the wild next week!

At the very least it's an improvement :) Thanks for your help here!

PR #228, opened 2024-10-09
Deprecation: Fix ptp --> np.ptp for numpy 2.0.0

CSSFrancis:
Support numpy 2.0.0

#### Progress of the PR
- [x] Pass tests with numpy 2.0.0

#### For reviewers

- [x] The PR title is short, concise, and will make sense 1 year later.
- [x] New functions are imported in corresponding `__init__.py`.
- [x] New features, API changes, and deprecations are mentioned in the
- [x] Contributor(s) are listed correctly in `credits` in `diffsims/release_info.py` and
      in `.zenodo.json`.

### DIFF
--- a/diffsims/tests/crystallography/test_get_hkl.py
+++ b/diffsims/tests/crystallography/test_get_hkl.py
@@ -27,11 +27,7 @@ from diffsims.crystallography import (
     get_hkl,
     ReciprocalLatticeVector,
 )
-
-if np.__version__ >= "2.0.0":
-    from numpy.exceptions import VisibleDeprecationWarning
-else:
-    VisibleDeprecationWarning = np.VisibleDeprecationWarning
+from diffsims.constants import VisibleDeprecationWarning
 
 
 class TestGetHKL:
--- a/diffsims/tests/crystallography/test_get_hkl.py
+++ b/diffsims/tests/crystallography/test_get_hkl.py
@@ -28,12 +28,17 @@ from diffsims.crystallography import (
     ReciprocalLatticeVector,
 )
 
+if np.__version__ >= "2.0.0":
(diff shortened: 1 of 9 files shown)


CSSFrancis:
@hakonanes If you want to review/merge this I can start the process of making a numpy 2.0.0 patch release.

diffsims/tests/crystallography/test_get_hkl.py
```diff
@@ -28,12 +28,17 @@
     ReciprocalLatticeVector,
 )
 
+if np.__version__ >= "2.0.0":
```
hakonanes: I suggest to add a #TODO comment stating that this check should be removed when NumPy >= 2 is the minimal supported version.

Should be done the other places as well.

diffsims/tests/utils/test_deprecation.py
```diff
@@ -5,6 +5,11 @@
 
 from diffsims.utils._deprecated import deprecated, deprecated_argument
 
+if np.__version__ >= "2.0.0":
```
hakonanes: Instead of repeating this three places, I suggest to add a private `diffsims.constants` module like we've done in orix: https://github.com/pyxem/orix/blob/develop/orix/constants.py. We can then import this warning `from diffsims.constants import VisibleDeprecationWarning`.

hakonanes (approved):
Two minor suggestions. Not required for merging.

diffsims/tests/crystallography/test_get_hkl.py
hakonanes: Actually, this should be >= "1.25.0" because that's when `numpy.exceptions.VisibleDeprecationWarning` was introduced.

Issue #234, opened 2025-01-23
Simulations with Phase do not take symmetry into account


viljarjf:
**Describe the bug**
The new simulations with a Phase does not (necessarily) take symmetry into account when calculating intensities. Only the explicitly included atoms in the underlying structure are part of the structure factor calculations.

**To Reproduce**
```python
from numpy import allclose
from matplotlib import pyplot as plt
from diffpy.structure import Structure, Atom, Lattice
from orix.crystal_map import Phase
from orix.quaternion import Rotation
from diffsims.generators.simulation_generator import SimulationGenerator

l = Lattice(10, 10, 10, 90, 90, 90)
a = [Atom("Au", (0, 0, 0))]
s = Structure(a, l)

P1 = Phase(structure=s, space_group=1)
Fd3m = Phase(structure=s, space_group=227)

# Rotate slightly off-zone to see extinct reflections
rot = Rotation.from_euler((0, 0, 1), degrees=True)

gen = SimulationGenerator()
P1_sim = gen.calculate_diffraction2d(P1, rotation=rot)
Fd3m_sim = gen.calculate_diffraction2d(Fd3m, rotation=rot)

# Use ReciprocalLatticeVector.sanitize_phase
Fd3m_sim.coordinates.sanitise_phase()
Fd3m_expanded = Fd3m_sim.coordinates.phase
Fd3m_sim_expanded = gen.calculate_diffraction2d(Fd3m_expanded, rotation=rot)

# Intensities should be different due to the added symmetry, 
# even when the symmetrically equivalent atoms are missing from the structure
assert not allclose(P1_sim.coordinates.intensity, Fd3m_sim_expanded.coordinates.intensity) # Passes
# assert not allclose(P1_sim.coordinates.intensity, Fd3m_sim.coordinates.intensity) # Fails

# Plot patterns
plt.figure()
plt.subplot(1, 3, 1)
plt.title("P1")
P1_sim.plot(ax=plt.gca())

plt.subplot(1, 3, 2)
plt.title("Fd3m default")
Fd3m_sim.plot(ax=plt.gca())

plt.subplot(1, 3, 3)
plt.title("Fd3m expanded")
Fd3m_sim_expanded.plot(ax=plt.gca())

```
![Image](https://github.com/user-attachments/assets/75be2612-59be-4db0-bbfb-50ef5d7be81d)

Currently, intensities are calculated as follows:
1. Find intersecting vectors for a given reflection
[...]

PR #238, opened 2025-05-30
Bugfix: Small Bugfix for polar flatten simulation

CSSFrancis:
It seems like we weren't doing a great job of masking values etc which when doing a orientation map could have some pretty big impacts. 

We should make a bug patch release probably asap. 

#### Progress of the PR

### DIFF
--- a/diffsims/tests/generators/test_simulation_generator.py
+++ b/diffsims/tests/generators/test_simulation_generator.py
@@ -19,6 +19,7 @@
 import numpy as np
 import pytest
 from pathlib import Path
+import re
 
 import diffpy.structure
 from orix.crystal_map import Phase
@@ -359,10 +360,9 @@ def test_calculate_diffraction2d_progressbar_single_phase(capsys):
     sims = gen.calculate_diffraction2d(phase, rots, show_progressbar=True)
 
     captured = capsys.readouterr()
-    expected = "test phase: 100%|██████████| 10/10"  # also some more, but that is compute-time dependent
-    # ignore possible flushing
-    captured = captured.err.split("\r")[-1]
-    assert "████" in captured  # This can be faulty on CI, but we check for the bar
+    # Accept any number of "█" characters
+    expected = "test phase: 100\%\|█+\| 10\/10"
+    assert re.findall(expected, captured.err)
 
 
 def test_calculate_diffraction2d_progressbar_multi_phase(capsys):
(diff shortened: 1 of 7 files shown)


CSSFrancis:
pre-commit.ci autofix

CSSFrancis:
@wiscott24 Does this fix your problem?

CSSFrancis:
@viljarjf any chance that you can look at this? This fixes the failing tests with #235.  Sorry it took me so long to get around to that.  Also feel free to just ping me for stuff if I'm being slow.  Sometimes I see things and mean to go back to it and forget :)

viljarjf (approved):
Seems good! There might be some merit to giving a warning if the phase is not symmetrically expanded, as the structure factors will not be correct otherwise. just something simple like
```python
if len(phase.structure) != len(phase.expand_asymmetric_unit().structure):
    warnings.warn(...)
```
Otherwise this looks fine to me! I have made some changes to calculate_simulation2d in another draft PR, I can see if the same can be applied to simulation1d when I get around to finalizing it.

diffsims/tests/generators/test_simulation_generator.py
```diff
@@ -363,7 +362,7 @@ def test_calculate_diffraction2d_progressbar_single_phase(capsys):
     expected = "test phase: 100%|██████████| 10/10"  # also some more, but that is compute-time dependent
     # ignore possible flushing
     captured = captured.err.split("\r")[-1]
-    assert captured[: len(expected)] == expected
+    assert "████" in captured  # This can be faulty on CI, but we check for the bar
```
viljarjf: These can be done with regex to allow any number of the block character. I'll give it a go later today

viljarjf: https://github.com/CSSFrancis/diffsims/pull/3

wiscott24:
> @wiscott24 Does this fix your problem?

Yes, thanks!

CSSFrancis:
@viljarjf hmmm. 

Should we just have something like:

```python
if len(phase.structure) != len(phase.expand_asymmetric_unit().structure):
    phase = phase.expand_asymmetric_unit()
```

viljarjf:
> @viljarjf hmmm.
> 
> Should we just have something like:
> 
> ```python
> if len(phase.structure) != len(phase.expand_asymmetric_unit().structure):
>     phase = phase.expand_asymmetric_unit()
> ```

I'm not really sure. Users might want to only have certain atoms, to see how that might affect things? If so, they can use P1 I guess. On the other hand, I expect most users use a cif, in which case all atoms are present already. Displaying a warning is also not done in many places, outside of maybe orix warning you if it thinks you meant to use `degrees=True`, so this might seem out of place. 

I say we can leave it for now, maybe move it to a seperate issue for further discussion. This is also a potential problem in 2D, ref #234

